PR #20, opened 2016-09-26
Unittesting added for bridge_common

GowthamShanmugam:
unit testing added for bridge common and contents of __init__ file is moved to a new file etcdobj to solve  log problem


review-ninja:
<a href="https://app.review.ninja/Tendrl/bridge_common/pull/20" target="_blank">ReviewNinja

bridge_common/tests/base.py
```diff
@@ -1,6 +1,5 @@
-from oslotest import base
+import unittest
```
r0h4n: We are using py.test which doesnt require you to inherit "unittest.TestCase"

please read more http://doc.pytest.org/

GowthamShanmugam: i will change it

r0h4n:
Seems like there's py.test related dependency issues, Check travis for more details and send out a new fix please

Issue #13, opened 2016-09-12
Definition YAML files should have flags to control the visibility of entities


brainfunked:
Both the state and the operations definition files should allow the visibility of all the entities (objects, atoms, attributes and flows) via the API, to be controlled. This enables a workflow where the supporting backend implementation would first have to implement the support for a certain entity. Once the implementation is complete, that entity would then be marked as available for manipulation via the API.

To do this, the definition file needs to support:

- a global default visibility flag, and

- per entity visibility flags that overrides the global default.

r0h4n:
Fixed https://github.com/Tendrl/bridge_common/blob/master/etc/samples/tendrl_definitions_generic-0.1.sample.yaml#L7

PR #34, opened 2016-10-24
Tendrl SDS/Node definitions Schema 0.1

r0h4n:


r0h4n:
@brainfunked @nthomas-redhat @anivargi @nnDarshan @shtripat @GowthamShanmugam @julienlim 
Please review and give inputs

- Now a node_agent or a sds will have only one yaml file "tendrl_definitions_<sds|node_agent>-nvr.yaml", All definitions (objects, attrs, atoms, flows) should be in this yaml.
- Please fix any PR's of yours which might be affected by this schema change.
- File bugs against current definitions of tendrl/ceph_bridge, tendrl/gluster_bridge, tendrl/node_agent if they dont match the schema defined in this PR

etc/samples/tendrl_definitions_generic-0.1.sample.yaml
```diff
@@ -0,0 +1,77 @@
+---
+valid_objects:
+- volume
+
+object_details:
+  volume:
+    enabled: True
```
shtripat: What this is meant for and based on what we should set True/False?
Is it like whether this feature is enabled in tendrl??

etc/samples/tendrl_definitions_generic-0.1.sample.yaml
```diff
@@ -0,0 +1,77 @@
...
+    value: "/clusters/gluster/$cluster.id/volumes/$volume.id
```
shtripat: Here gluster doesnt maintain any UUID for cluster but does for volumes and bricks etc. Is it like responsibility of the bridge code to decide how to populate the UUID for an object?

etc/samples/tendrl_definitions_generic-0.1.sample.yaml
```diff
@@ -0,0 +1,77 @@
...
+        type: Integer
+      transport:
+        type: String
+      brickdetails:
+        type: Brick[]
+      force:
+        type: Boolean
+      directory:
+        type: String
+      bitrot:
```
shtripat: Fields specific to bitrot, quota etc might not be required to be maintained at the moment. They could be added as and when features get added to tendrl I feel. What other think on this ?

etc/samples/tendrl_definitions_generic-0.1.sample.yaml
```diff
@@ -0,0 +1,77 @@
+---
```
shtripat: I would suggest another PR with sample values for CreateVolume flow may be. That would make lives easier for other PRs as few of the fields might be understood well within this PR.

r0h4n: I will make that happen. But please raise questions about any items here

etc/samples/tendrl_definitions_generic-0.1.sample.yaml
r0h4n: Yes, it is for all Tendrl components to understand if the feature is enabled or not. This is for testing and staging purpose fixes #13

etc/samples/sds_objects_generic.sample.yaml
```diff
@@ -0,0 +1,37 @@
+# This schema is not used stand-alone, it is a sub-section of "tendrl_definitions_<sds_nvr>.yaml" | "tendrl_definitions_generic.sample.yaml
+
+---
+valid_objects:
+- volume
+
+object_details:
+  volume:
+    enabled: True
+    value: "/clusters/gluster/$cluster.id/volumes/$volume.id
```
anivargi: @r0h4n we cannot have that schema for the etcd value this way, to get the details I need to know the sds_type (gluster in this example), which is not possible to know till I get the cluster details. I would prefer /clusters/$cluster.id/volumes/$volume.id with the sds_type and sds_version inside the cluster details

etc/samples/tendrl_definitions_generic-0.1.sample.yaml
shtripat: Thanks. That is required sooner as other PRs need to update based on that reference one.

etc/samples/tendrl_definitions_generic-0.1.sample.yaml
shtripat: ok

etc/samples/tendrl_definitions_generic-0.1.sample.yaml
r0h4n: Sure

etc/samples/tendrl_definitions_generic-0.1.sample.yaml
```diff
@@ -0,0 +1,77 @@
...
+      scrub-frequency:
+        type: Integer
+      scrub-throttle:
+        type: Integer
+    atoms:
+        create:
+          enabled: True
+          uuid: statically generated per atom
+          # List of flows this of which this atom can be part of
+          flows: ["CreateGlusterVolume"]
```
shtripat: Why do we have reference of flows in atoms and reverse of this in flows (reference of atoms in flows)?
Looks A --> B and B --> A relationship. Do we need definition for both of these in specs?

etc/samples/sds_objects_generic.sample.yaml
r0h4n: That makes sense. But in reference to @shtripat comments about uuid above, I would prefer mapping the tendrl world cluster uuid's to their sds_type and other details at 

eg: assume 1 gluster cluster and 1 ceph cluster already created/imported
- /clusters/gluster/12321321/
- /clusters/ceph/32431432/

You can then assume to find mappings of above tendrl data at 
- /clusters/tendrl_context: {"12321321" : {sds_type: "gluster-<nvr>", rest of the metadata},  "32431432": {"sds_type": "ceph-<nvr>", rest of the metadata}

works for you?

etc/samples/tendrl_definitions_generic-0.1.sample.yaml
r0h4n: IMHO, the schema allows to update other PRs, You can go ahead and make changes. We can all review your changes, you dont need to be blocked for a sample at this stage of our development

etc/samples/tendrl_definitions_generic-0.1.sample.yaml
r0h4n: Yes, these references can be removed later if we dont need them, required for now. Doesnt impact much in our processing of these entities

shtripat: Ok. To me its looks redundant actually. For what purpose we need this at the moment?

etc/samples/sds_objects_generic.sample.yaml
r0h4n: Check latest patch, have tried out @anivargi 's comments

etc/samples/tendrl_definitions_generic-0.1.sample.yaml
```diff
@@ -1,11 +1,19 @@
...
+    enabled: True
+    value: /clusters/$cluster.id/tendrl_context
+    attrs:
+      sds_type:
+        type: String
+        help: "NVR of the Tendrl managed sds, eg: 'gluster-3.2.0'"
   volume:
     enabled: True
-    value: "/clusters/gluster/$cluster.id/volumes/$volume.id
+    value: "/clusters/$cluster.id/volumes/$volume.id
```
shtripat: Looks fine to me. @anivargi ?

etc/samples/tendrl_definitions_generic-0.1.sample.yaml
r0h4n: tendrl-api exposes objects and the atoms valid for those objects, there needs to be a way we map these atoms to flows (eg: create gluster volume atom can be run via createglustervolume flow or the createglustervolumewithoutbricks flow) based on tendrl-api job inputs, hence required for now.

etc/samples/tendrl_definitions_generic-0.1.sample.yaml
```diff
@@ -0,0 +1,85 @@
...
+      description: "Create Gluster Volume with given context/params"
+      atoms: ["volume.atoms.create"]
+      pre_run : ["tendrl.node_agent.atoms.<atom_name>.<boolean_function>1"...]
+                #boolean returning pre/post conditions function list
+      post_run: ["tendrl.node_agent.atoms.<atom_name>.<boolean_function>1"...]
+      run: "tendrl.gluster_bridge.flows.create_gluster_volume.CreateGlusterVolume"
+      type: Create
+      outputs:
+      - "A map of type {object.attr1 : {required: True}, object
+     .attr2:{required: False}}}} that this will be providing (or could
```
nnDarshan: Is "{object.attr1 : {required: True}, object .attr2:{required: False}}}} for "inputs" attribute ? seems confusing to me here

etc/samples/atoms_generic.sample.yaml
```diff
@@ -0,0 +1,26 @@
...
+          enabled: True
+          uuid: statically generated per atom
+          name: "Name"
+          # List of flows this of which this atom can be part of
+          flows: ["TendrlFlow1"]
+          run: "tendrl.node_agent.atoms.cpu.create.Create"
+          type: "get"
+          pre_run : ["tendrl.node_agent.atoms.<atom_name>.<boolean_function>1"...]
+          post_run: ["tendrl.node_agent.atoms.<atom_name>.<boolean_function>1"...]
+          name: Meaningful name for this atom, should be something that is distinguishable and understandable for notification, debugging, storing and any other similar purposes
```
nnDarshan: Please remove this

etc/samples/tendrl_definitions_generic-0.1.sample.yaml
r0h4n: "outputs" is a dict of results provided on completion of execution of the flow, these can be diverse objects. 

"inputs" is a dict of inputs gathered from the tendrl-api-job which can be used by the flow and atoms in the flow

makes sense? Basically dicts of required input for the flow and dict of output returned by the flow

etc/samples/atoms_generic.sample.yaml
```diff
@@ -0,0 +1,26 @@
...
+    atoms:
+        create:
+          enabled: True
+          uuid: statically generated per atom
+          name: "Name"
+          # List of flows this of which this atom can be part of
+          flows: ["TendrlFlow1"]
+          run: "tendrl.node_agent.atoms.cpu.create.Create"
+          type: "get"
+          pre_run : ["tendrl.node_agent.atoms.<atom_name>.<boolean_function>1"...]
```
nthomas-redhat: I assume this is again a list of functions. Is there any order in which they run?

etc/samples/atoms_generic.sample.yaml
```diff
@@ -0,0 +1,26 @@
...
+          post_run: ["tendrl.node_agent.atoms.<atom_name>.<boolean_function>1"...]
+          name: Meaningful name for this atom, should be something that is distinguishable and understandable for notification, debugging, storing and any other similar purposes
+          outputs:
+          - "A map of type {object.attr1 : {required: True}, object.attr2:{required: False}}}} that this will be providing (or could provide) to others, used to correlate and associate the thing/s this atom produces, if it produces anything at all."
+
+          # Required input object.attribute list
+          inputs:
```
nthomas-redhat: Is it map or list? comment says its a list

etc/samples/atoms_generic.sample.yaml
nthomas-redhat: I think its good add some doc around each attributes here , what each of that means and if there any conditions associated with it. I agree that most of those convey the meaning but good to have

etc/samples/atoms_generic.sample.yaml
```diff
@@ -0,0 +1,26 @@
+# This schema is not used stand-alone, it is a sub-section of "tendrl_definitions_<sds_nvr>.yaml" | "tendrl_definitions_generic.sample.yaml
+
+---
+sds_version: gluster-3.8.3
+object_details:
+  generic_object:
+    atoms:
+        create:
+          enabled: True
+          uuid: statically generated per atom
```
nthomas-redhat: In case if they clash, what are the after effects. How we come to know that there is a clash? Some how if we know that there is a clash(parsing?) then we are good

etc/samples/atoms_generic.sample.yaml
r0h4n: They are unordered as of now

etc/samples/atoms_generic.sample.yaml
r0h4n: Its a dict

r0h4n:
@brainfunked @nthomas-redhat Please approve this PR

etc/samples/flows_generic.sample.yaml
```diff
@@ -0,0 +1,28 @@
...
+      post_run: ["tendrl.node_agent.atoms.<atom_name>.<boolean_function>1"...]
+      run: "tendrl.node_agent.flows.create_ceph_cluster"
+      type: Create
+      outputs:
+      - "A map of type {object.attr1 : {required: True}, object
+     .attr2:{required: False}}}} that this will be providing (or could
+     provide) to others, used to correlate and associate the thing/s this flow
+      produces, if it produces anything at all."
+      inputs:
+      - An immutable {object1.attr1 : {value: "cluster1"}, object
```
TimothyAsir: Any reason why this input, output tags provided here (in flow) and in atoms also?
Which input should we use for validation? (from atom / flow)

etc/samples/tendrl_definitions_generic-0.1.sample.yaml
```diff
@@ -0,0 +1,77 @@
...
+          version: An immutable version that associates version information with this atom. It can be useful in resuming older versions of atoms. Standard major, minor versioning concepts should apply
+
+
+  flows:
+    CreateGlusterVolume:
+      enabled: True
+      tendrl_context: "A map of Tendrl flow level context items like cluster id,tendrl id etc"
+      uuid: "pre-generated uuid here"
+      description: "Create Gluster Volume with given context/params"
+      atoms: ["volume.atoms.create"]
```
TimothyAsir: Need one small clarification here!
if atoms has multiple atom for a particular flow (ex:- atoms: ['volume.atoms.create', 'volume.atoms.start'] ), then we may need to validate job api inputs against all the atoms right?

etc/samples/flows_generic.sample.yaml
r0h4n: Flow and atoms are separate entities, please validate them separately. 

The reason we have inputs and outputs for both flows and atoms is so that we can have flow level data and atom level data.

r0h4n:
@brainfunked Please review, need to merge soon

etc/samples/atoms_generic.sample.yaml
shtripat: I assume inputs would be defined as below 

inputs: ['volname', 'stripe_count', ....]

Now with this how to mention in inputs if a field is optional or mandatory?

shtripat: I am bit confused here I feel because volume creation we can pass stripe_count if we want a striped volume, else with only "volname" and "brickdetails" it would create a distributed volume by default.
How to handle this kind of scenario?

PR #38, opened 2016-10-27
Added mandatory and optional fields for inputs

shtripat:
Added two different lists for inputs attributes to receive
mandatory and optional arguments for the flows and atoms.


shtripat:
@r0h4n @nthomas-redhat added sample changes as yaml changing.

PR #43, opened 2016-10-27
Update tendrl_definitions_generic-0.1.sample.yaml

r0h4n:


etc/samples/tendrl_definitions_generic-0.1.sample.yaml
```diff
@@ -1,84 +1,85 @@
...
+      mandatory: 
+        - "A unique set/list of mandatory attribute names [object1.attribute1...] for ensuring valid input attributes are provided in the incoming tendrl api job"
+      optional: 
+        - "A unique set/list of optional attribute names [object1.attribute1...]. If the values are passed they would be validated for types and valid values."
+    outputs: 
+      - "A map of type {object.attr1 : {required: True}, object .attr2:{required: False}}}} that this will be providing (or could provide) to others, used to correlate and associate the thing/s this flow produces, if it produces anything at all."
+    post_run: 
+      - tendrl.gluster_bridge.atoms.<atom_name>.<boolean_function>1...
+    pre_run: 
+      - tendrl.gluster_bridge.atoms.<atom_name>.<boolean_function>1...
```
shtripat: @r0h4n @brainfunked @nthomas-redhat 
As per my understanding we can have validation routines written under say atoms/volume/validations.py and those functions could be referred here.
Having atom's input only passed here might not suffice and would always need cluster-id as well to be passed for all the atoms.
Should we go ahead and add cluster-id as well under inputs for all the volume operations in that case?

PR #50, opened 2016-11-09
Initialize logger in common module as well

shtripat:


shtripat:
@r0h4n @nthomas-redhat @brainfunked please review

PR #51, opened 2016-11-09
Tendrl SDS/Node definitions Schema 0.1

r0h4n:
- Remove "tendrl_context" as first level object
  It should be collected via "inputs" of flows and atoms


r0h4n:
@nnDarshan @nthomas-redhat

PR #64, opened 2016-11-21
common: Add singleton utility

anmolbabu:
This patch adds a utility func called singleton if any
class is decorated as singleton, such a class will
behave equivalent to java singleton.

Classes such as those to read configuration can be made
singleton and created only once and hence configuration
is read only once and creating instance of such a class is then
guranteed to be same as what is created the first time. A
provision can then be made within this method to force read
the config file if required for supporting cases like service reload
(instead of restart). This can also be used to create etcd instance
once and use wherever required as that avoids having to explicitly
parse the same arguments from same source repeatedly . 
However, currently with time series db client wrapper being plugin
based, its used only for accessing the time series db plugin as per
configuration in tendrl conf.

tendrl-bug-id: Tendrl/common#69
Change-Id: Ibf86424ee4f5718965cae3cae89cc788f301a920


tendrl/common/singleton.py
```diff
@@ -0,0 +1,8 @@
+def singleton(cls):
```
shtripat: It would be good if you add usage as well in commit message.
Looks good to me.

r0h4n: Please change method name to something more meaningful, eg: to_singleton, make_singleton

anmolbabu: @shtripat done @r0h4n done

r0h4n:
Please format the commit msg as per the example https://github.com/Tendrl/gluster_integration/commit/119e98fa1763356d17b1e429a8a16d9be0624044

PR #77, opened 2016-12-08
Makefile: safer "make clean" operations

ktdreyer:
`make clean` deletes single files. There's no need for the "-r" (recursive) flag, and it's dangerous to run rm recursively unless absolutely necessary.


ktdreyer:
@r0h4n I've rebased  to the tip of master

PR #79, opened 2016-12-09
[WIP]Common: Added generic utils to common

nnDarshan:
This patch adds generic utils like command-executor,
package-installer, service-manager to common as utility.


nnDarshan:
@r0h4n @nthomas-redhat @shtripat @brainfunked Its initial WIP patch for adding general utilities.

@shtripat Can you please help me by providing an issue for this, when you create one for the re factoring specification.

nnDarshan:
@r0h4n @nthomas-redhat @shtripat Please review.

nnDarshan:
@shtripat If you find it easy to use these utils as part of your re-factoring, please go ahead. If you find it would make your re-factoring patch huge , I will send a separate PR afterwards to do that. In any case, please provide your thoughts

shtripat:
@nnDarshan I feel doing this as part of different PR is good as refactoring PR is already huge and I am targetting more of cleanup, movement to common module and proper class hierarchies etc as part of my PR.

@r0h4n @brainfunked @nthomas-redhat thoughts??

nnDarshan:
tendrl-spec: Tendrl/specifications#31
tendrl-bug-id: Tendrl/common#80

tendrl/common/utils/service.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def reload(self):
+        attr = self.attributes
+        attr.update({"state": "reloaded"})
+        return self.__run_module(attr)
+
+    def restart(self):
+        attr = self.attributes
+        attr.update({"state": "restarted"})
+        return self.__run_module(attr)
```
shtripat: Is it possible to add a status() function as well. Few modules need that I feel.

nnDarshan: The ansible module which we are using to manage the service has only 4 types of operations namely start, stop, restart and reload(http://docs.ansible.com/ansible/service_module.html refer state under options). So its not possible with this approach.
I feel when we monitor systemd dbus signals we will be able to obtain the service states

tendrl-common.spec
```diff
@@ -7,13 +7,15 @@ Source0: %{name}-%{version}.tar.gz
 License: LGPLv2+
 URL: https://github.com/Tendrl/common
 
+BuildRequires: ansible
```
shtripat: I remember on rhel by default ansible 2.2 was not getting downloaded. We should explicitly mention > 2.2 here if you use any ansible 2.2 specific feature.

tendrl/common/tests/test_ansible_runner.py
```diff
@@ -0,0 +1,75 @@
...
+from ansible import modules
+import os
+import pytest
+from tendrl.common.utils.ansible_module_runner \
+    import AnsibleExecutableGenerationFailed
+from tendrl.common.utils.ansible_module_runner \
+    import AnsibleRunner
+
+
+class Test_ansible_runner_constructor(object):
```
shtripat: Better use CamelCase for all the class names

tendrl/common/utils/install.py
```diff
@@ -0,0 +1,38 @@
+from ansible_module_runner import AnsibleExecutableGenerationFailed
```
shtripat: I feel source could be named as package_installer.py

tendrl/common/utils/service.py
shtripat: ok

tendrl-common.spec
nnDarshan: Done

tendrl/common/utils/install.py
nnDarshan: Done

tendrl/common/utils/command.py
```diff
@@ -0,0 +1,24 @@
...
+from ansible_module_runner import AnsibleRunner
+import logging
+
+ANSIBLE_MODULE_PATH = "core/commands/command.py"
+LOG = logging.getLogger(__name__)
+
+
+class Command(object):
+    def __init__(self, command):
+        self.attributes = {"_raw_params": command}
```
r0h4n: This needs a whitelist of good commands we use or some sorta security check

nnDarshan:
@r0h4n @nthomas-redhat @brainfunked Please Review. Have addressed the commets

tendrl/common/utils/command.py
```diff
@@ -5,9 +5,25 @@
 ANSIBLE_MODULE_PATH = "core/commands/command.py"
 LOG = logging.getLogger(__name__)
 
+SAFE_COMMAND_LIST = [
```
shtripat: Should we externalize these command in tendrl owned yaml file? Just a thought, but if we are very strict on list of commands supported, then it ok to keep inside code. My only point is just for a new command we would need a code change later.

Comment??

nnDarshan: My opinion is having it in code is better, keeping the security in mind. 
@nthomas-redhat @r0h4n @brainfunked  Please provide your thoughts

tendrl/common/utils/service.py
```diff
@@ -0,0 +1,63 @@
+from ansible_module_runner import AnsibleExecutableGenerationFailed
+from ansible_module_runner import AnsibleRunner
+import logging
+
+ANSIBLE_MODULE_PATH = "core/system/systemd.py"
```
nthomas-redhat: This supports only systemd services only right?
If we want to support old versions of LINUX where systemd is not available(ex;RHEL6), this would not be enough. I think  the service module(https://github.com/ansible/ansible-modules-core/blob/devel/system/service.py) has support for different init modules including systemd and BSD init. Have considered that?

tendrl/common/utils/service.py
nthomas-redhat: start and stop are idempotent actions that will not run commands unless necessary, so you can rely on these itself and you get the status in return. Can we take that approach?

nnDarshan: But using this approach we can't just read the status without changing it right ? If we already know that we want the state to be either started or stopped we can directly use the start/stopped method. I am not getting the advantage of using a status that way. Could you please elaborate ?

tendrl/common/utils/command.py
shtripat: ok

tendrl/common/utils/service.py
nnDarshan: Done

PR #87, opened 2016-12-28
Add alert model and a util to assist alert operations

anmolbabu:
tendrl-bug-id: Tendrl/alerting#10,Tendrl/alerting#11,Tendrl/alerting#12,Tendrl/alerting#14

### DIFF
--- a/tendrl/common/alert.py
+++ b/tendrl/common/alert.py
@@ -1,0 +1,109 @@
+import etcd
+import json
+from tendrl.common.config import TendrlConfig
+from tendrl.common.singleton import to_singleton
+
+alert_severity_map = {
+    'INFO': 0,
+    'WARNING': 1,
+    'CRITICAL': 2
+}
+
+
+class Alert(object):
+    def __init__(self, alert_id, node_id, time_stamp, resource, current_value,
+                 tags, alert_type, severity, significance, ackedby, acked,
+                 pid, source):
+        self.alert_id = alert_id
+        self.node_id = node_id
+        self.time_stamp = time_stamp
+        self.resource = resource
+        self.current_value = current_value
+        self.tags = tags
+        self.alert_type = alert_type
+        self.severity = severity
+        self.significance = significance
+        self.ackedby = ackedby
+        self.acked = acked
+        self.pid = pid
+        self.source = source
+
+    def to_json_string(self):
+        return json.dumps(self.__dict__)
+
+    def is_same(self, alert2):
+        if self.resource != alert2.resource:
+            return False
+        if self.alert_type != alert2.alert_type:
+            return False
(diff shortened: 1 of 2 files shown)


anmolbabu:
TODO(@anmolbabu) Add UTs

shtripat:
@anmolbabu tendrl-bug-id twice in commit message. correct the same

tendrl/common/alert.py
```diff
@@ -0,0 +1,109 @@
...
+        if 'current_value' not in alert:
+            raise KeyError('current_value')
+        if 'alert_type' not in alert:
+            raise KeyError('alert_type')
+        return alert
+
+    def store_alert(self, alert):
+        self.etcd_client.write(
+            '/alerts/%s' % alert.alert_id,
+            alert.to_json_string()
```
shtripat: You want json representation of class Alert to be saved in etcd or mark Alert class itself as a EtcdObject??

anmolbabu: Marking Alert as EtcdObject doesn't work because EtcdObject requires fields to be class variables and not instance variables. The problem with this is that, there can be multiple alerts at same time so this static kind of fields don't suit our requirement.

anmolbabu:

tendrl/common/alert.py
shtripat: Got it. Its ok

tendrl/common/alert.py
```diff
@@ -0,0 +1,153 @@
...
+        self.source = source
+
+    def to_json_string(self):
+        return json.dumps(self.__dict__)
+
+
+@to_singleton
+class AlertUtils(object):
+    def __init__(self):
+        etcd_kwargs = {
```
shtripat: I feel config should come from AlertManager and we should not refer create a local etcd_client. If possible pass from AlertManager.

anmolbabu: ok

tendrl/common/alert.py
r0h4n: @anmolbabu could you please track a future spec which will eliminate this need of having class variables as EtcdObject? this would require changes to the commons/etcdobj

tendrl/common/alert.py
```diff
@@ -0,0 +1,153 @@
...
+
+alert_severity_map = {
+    'INFO': 0,
+    'WARNING': 1,
+    'CRITICAL': 2
+}
+config = TendrlConfig("commons", "/etc/tendrl/tendrl.conf")
+
+
+class Alert(object):
```
r0h4n: This needs to defined via Tendrl definitions under say the alerting namepsace or node-agent namespace, we dont use standalone objects

r0h4n: You can do that in a separate spec/bug, for now it is ok

PR #92, opened 2017-01-05
Enhanced to look for node_ids parameter in job

shtripat:
Now the job processor looks for a parameter with
name `node_ids` and if current node's id falls
under the same list, it would pick and process the
job. Only one of the nodes from the list would pick
the job for processing.

tendrl-bug-id: Tendrl/common#86
tendrl-spec: multiple_node_ids_in_jobs.adoc

### DIFF
--- a/etc/samples/tendrl_api_job.sample.json
+++ b/etc/samples/tendrl_api_job.sample.json
@@ -10,6 +10,7 @@
 	"cluster_id": "bc35fde8-528b-4908-8d52-50d8bf964dd4",
 	"sds_name": "gluster",
 	"sds_version": "3.2.0",
+	"node_ids": ["node_id1", ...],
 	"run": "tendrl.gluster_integration.flows.create_volume.CreateVolume",
 	"status": "new",
 	"parameters": {
@@ -24,7 +25,7 @@
 
 {
 	"cluster_id": "bc35fde8-528b-4908-8d52-50d8bf964dd4",
-	"node_id": "node_id", 
+	"node_ids": ["node_id1", ...],
 	"run": "tendrl.node_agent.gluster_integration.flows.import_cluster.ImportCluster",
 	"status": "new",
 	"parameters": {
--- a/tendrl/common/manager/manager.py
+++ b/tendrl/common/manager/manager.py
@@ -33,6 +33,7 @@ class Manager(object):
         self,
         name,
         integration_id,
+        node_id,
         config,
         events,
         persister_thread,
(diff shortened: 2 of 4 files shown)


shtripat:
@Tendrl/tendrl-core please review.

GowthamShanmugam:
@shtripat LGTM

shtripat:
@r0h4n we need to merge the spec PR https://github.com/Tendrl/specifications/pull/104 first and then only this can be merged.

etc/samples/tendrl_api_job.sample.json
```diff
@@ -24,7 +25,7 @@
 
 {
 	"cluster_id": "bc35fde8-528b-4908-8d52-50d8bf964dd4",
-	"node_id": "node_id", 
+	"node_ids": ["node_id1", ...],
```
r0h4n: rename this as "target_node_ids"

tendrl/common/manager/manager.py
```diff
@@ -33,6 +33,7 @@ def __init__(
         self,
         name,
         integration_id,
+        node_id,
```
r0h4n: Hold on, would this mean that even the gluster_integration and ceph_integration processes will know about the node_id? This isnt require

@brainfunked

tendrl/common/manager/rpc_job_process.py
```diff
@@ -57,6 +57,12 @@ def _acceptor(self, job_type):
                     continue
                 raw_job = json.loads(job.value.decode('utf-8'))
                 try:
+                    # If current node's id does not fall under job's node_ids,
+                    # ignore the job and dont process
+                    if "node_ids" in raw_job:
```
r0h4n: this would mean there be "node_ids" in even jobs intended for the integration processes (gluster, ceph), why do you want that?

etc/samples/tendrl_api_job.sample.json
shtripat: done

tendrl/common/manager/manager.py
shtripat: So the Base manager class is common and node_agent, gluster_integration and ceph_integration managers extend from the same. Also the rpc job processor is common code.

The parameter node_ids is not a mandatory one. If its passed, current node_id is checked against the same. We dont have module specific rpc job processors anymore after refactoring, so this is required and I dont see an issue. Even in case of gluster and ceph integration, if required jobs can be targeted to specific set of nodes (like to mons in case of ceph cluster).

tendrl/common/manager/rpc_job_process.py
shtripat: Its not a mandatory parameter and if its present then we check if current node's id falls under this list.

PR #95, opened 2017-01-06
Modified flow extractor in rpc_job_process

nnDarshan:
As of now the extractor can parse the flow only if it
is in global namespace, this has been modified to handle
flows that are tied to object.

tendrl-spec: segragate_object_specific_flows
tendrl-bug-id: Tendrl/common#94


nnDarshan:

Issue #90, opened 2017-01-04
version in setup.py is not aligned with git tags


mbukatov:
Version specified in `setup.py` file is `0.0.1`, while the latest git tag (and releases page provided for these tags by github) is `v1.1`.

## Details

```
$ grep version setup.py
    version="0.0.1",
$ git tag
v1.0
v1.1
```

## Related Upstream Guidelines

https://packaging.python.org/distributing/#choosing-a-versioning-scheme
https://packaging.python.org/single_source_version/#single-sourcing-the-version

mbukatov:
Release process is fixed since git tag [v1.2](https://github.com/Tendrl/commons/releases/tag/v1.2). Since we can't fix this issue for already published tags mentioned above (`v1.0`, `v1.1`) as it would break history of public git repository, I consider this issue fixed.

PR #100, opened 2017-01-10
Refactor and decouple commons from config/context etc

r0h4n:
- Remove redundant samples, code

tendrl-bug-id: Tendrl/commons#90


tendrl/commons/flows/base_flow.py
```diff
@@ -39,17 +38,10 @@ def __init__(
...
 
+        # Add etcd_client to parameters via child flows
         self.parameters.update({'log': []})
-        etcd_kwargs = {
-            'port': int(self.config.get("commons", "etcd_port")),
-            'host': self.config.get("commons", "etcd_connection")
-        }
-
-        self.etcd_client = etcd.Client(**etcd_kwargs)
-        self.parameters.update({'etcd_client': self.etcd_client})
```
shtripat: I remember we had done this for a specific reason where we needed etcd_client in further jobs created while import cluster flow on other nodes. How we are planning to achieve this now?

tendrl/commons/manager/rpc_job_process.py
```diff
@@ -3,26 +3,21 @@
...
-    def __init__(self, syncJobThread):
-        self.config = syncJobThread._manager._config
-        etcd_kwargs = {
-            'port': int(self.config.get("commons", "etcd_port")),
-            'host': self.config.get("commons", "etcd_connection")
-        }
-
-        self.client = etcd.Client(**etcd_kwargs)
-        self.syncJobThread = syncJobThread
+    def __init__(self, syncjobthread, etcd_client):
```
shtripat: Below, the creation code for EtcdRpc is not passing etcd_client to the constructor.

tendrl/commons/persistence/etcd_persister.py
```diff
@@ -1,14 +1,8 @@
-from tendrl.commons.etcdobj.etcdobj import Server as etcd_server
 from tendrl.commons.persistence.persister import Persister
 
 
 class EtcdPersister(Persister):
-    def __init(self, config):
-        super(EtcdPersister, self).__init__(config)
```
shtripat: I feel passing config was more flexible option and later if we node something more from config, we would have to change the signature of constructor again to pass additional data.

tendrl/commons/flows/base_flow.py
r0h4n: creating an etcd_client is costly (default creates a httpconnectionpool of 10), hence it is ideal that the rpc layer provides an instance of the etcd_client by appending it to parameters during flow invocation

tendrl/commons/manager/rpc_job_process.py
r0h4n: Ack, will fix

tendrl/commons/persistence/etcd_persister.py
r0h4n: The persister and its methods should be atomic in nature, eg: save_volume(volume) should only save the volume to etcd and disk.

The reason for not sending a config to a persister is to keep it sane and simple :)

tendrl/commons/persistence/etcd_persister.py
```diff
@@ -1,14 +1,8 @@
...
-    def get_store(self):
-        etcd_kwargs = {
-            'port': int(self._config.get("commons", "etcd_port")),
-            'host': self._config.get("commons", "etcd_connection")
-        }
-        return etcd_server(etcd_kwargs=etcd_kwargs)
+    def __init(self, etcd_client):
+        super(EtcdPersister, self).__init__()
+        # Child classes responsible to provide valid etcd client
+        self.etcd_client = etcd_client
```
r0h4n: This should also be injected by the child class, will remove

tendrl/commons/flows/base_flow.py
shtripat: ok

PR #109, opened 2017-01-16
Fixed issues related to etcd_server parameter

shtripat:
tendrl-bug-id: Tendrl/commons#114

### DIFF
--- a/tendrl/commons/flows/base_flow.py
+++ b/tendrl/commons/flows/base_flow.py
@@ -156,18 +156,19 @@ class BaseFlow(object):
         namespace = atom_name.split('.objects.')[0]
         object_name = atom_name.split('.objects.')[1].split('.atoms.')[0]
         atoms = self.definitions[namespace]['objects'][object_name]['atoms']
-        atom = atoms[atom_name.split('.')[-1]]
+        atom = atoms[atom_name.split('.')[-2]]
         return atom.get('name'), atom.get('enabled'), atom.get('help'), \
-            atom.get('inputs'), atom.get('outputs'), atom.get('uuid')
+            atom.get('inputs'), atom.get('outputs'), atom.get('uuid'), \
+            atom.get('run')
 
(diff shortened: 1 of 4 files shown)


tendrl/commons/flows/base_flow.py
```diff
@@ -156,18 +156,19 @@ def extract_atom_details(self, atom_name):
         namespace = atom_name.split('.objects.')[0]
         object_name = atom_name.split('.objects.')[1].split('.atoms.')[0]
         atoms = self.definitions[namespace]['objects'][object_name]['atoms']
-        atom = atoms[atom_name.split('.')[-1]]
+        atom = atoms[atom_name.split('.')[-2]]
```
anmolbabu: In https://github.com/anmolbabu/performance_monitoring/blob/bc6e351e2f268aaf1c18d770a3fcf367ad33fe2c/tendrl/performance_monitoring/manager/tendrl_definitions_performance_monitoring.py I didn't require this change as I changed my definition file itself to suit this. So in my definition file I changed the atom names under the objects to class names instead of file names because as per the split in above line it turns out to be looking for the last part of the module path which is the class name and I verified the change to be working.please suggest what is the decided/better approach so that we can make changes accordingly

shtripat: I am ok with anything. The point is we are mixing uppercase and lowercase a lot for object and atom names. I personally feel we should restrict to uppercase for only object names and keep atom names lowercase only and as in above fix we are actually getting the module name for the atom details to be loaded from definitions.
@r0h4n @brainfunked @nthomas-redhat comments?

r0h4n: Please check https://github.com/Tendrl/node-agent/pull/147 for final naming. The problem needs to be tackled at both the source code side and the Definitions side.

The above PR will clear the confusion, keep an eye

r0h4n:
PR not relevant any more

PR #115, opened 2017-01-17
- Adds flow utils for job status updates

r0h4n:
- Changes etcd_client to etcd_orm
- etcd client can be found at "etcd_orm.client"


tendrl/commons/flows/base_flow.py
```diff
@@ -54,17 +56,8 @@ def run(self):
         # TODO(rohan) make a func outta the below logging common code
         msg = "Processing pre-runs for flow: %s" % self.job['run']
         LOG.info(msg)
```
shtripat: Line no 67, 88, 100, 120 are setting the log messages but they are not written immediately to etcd. Why so?

PR #121, opened 2017-01-20
Fix issues in commons

nnDarshan:
This patch fixes some basic issues in commons


nnDarshan:

shtripat:
Isn't it same change which https://github.com/Tendrl/commons/pull/109 does?
@r0h4n can we get these kind of critical fixes in so that we don't keep sending same changes. Even I have a PR which duplicates few stuff from one of @nnDarshan PRs.

PR #127, opened 2017-01-23
- Fix wrong use of namespace package's __init__ ( cannot have code in…

r0h4n:
… root __init__ https://www.python.org/dev/peps/pep-0420/#rationale)

tendrl-bug-id: Tendrl/specifications#37
tendrl-bug-id: Tendrl/specifications#54
tendrl-bug-id: Tendrl/specifications#32


r0h4n:
@nthomas-redhat @shtripat @nnDarshan @anmolbabu Please note

PR #134, opened 2017-01-24
Logging implementation for tendrl

GowthamShanmugam:

tendrl-spec: logging_implementation


tendrl/commons/message.py
```diff
@@ -0,0 +1,47 @@
...
+    def validate(self):
+        if (self.priority == "" or
+            self.timestamp == "" or
+            self.publisher == "" or
+            "message" not in self.payload):
+            return False
+        if "job_id" in self.payload:
+            if ("flow_id" not in self.payload or
+                "cluster_id" not in self.payload):
+                return False
```
GowthamShanmugam: @brainfunked @shtripat @nthomas-redhat @nnDarshan @anmolbabu 
For job cluster_id is mandatory or not? because node related jobs have cluster id?

import cluster job does not have cluster id right?

GowthamShanmugam:
@brainfunked @r0h4n @shtripat @nnDarshan @anmolbabu @nthomas-redhat please review

tendrl/commons/manager/__init__.py
```diff
@@ -12,27 +12,32 @@
 class Manager(object):
     def __init__(
             self,
+            message_handler_thread,
             sds_sync_thread,
-            central_store_thread,
```
brainfunked: The best practice is to keep the comma at the end for lists, if the last list element allows it. This gets people in the habit of adding a comma at the end of every list element, regardless of whether it's to be the last and prevent syntax errors.

tendrl/commons/event.py
```diff
@@ -0,0 +1,10 @@
+import socket
+import json
+
+
+class Event(object):
+    def __init__(self, message, host, port):
```
brainfunked: Supposed to be a UNIX socket. Refer to this comment and make the appropriate changes in both codebases: https://github.com/Tendrl/node-agent/pull/67/files#r97720923

tendrl/commons/message.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+class Message(object):
+    def __init__(self, message):
+        try:
+            if "message_id" not in message:
+                self.message_id = str(uuid.uuid4())
+            else:
+                self.message_id = message["message_id"]
+            self.timestamp = message["timestamp"]
```
anmolsachan: Shouldn't be timestamp be automatically generated when the object is instantiated? Otherwise, we will have to import datetime in every module

tendrl/commons/event.py
```diff
@@ -0,0 +1,10 @@
...
+        s = socket.socket()
+        s.connect((host, port))
+        s.sendall(json.dumps(message.__dict__))
+        s.shutdown(socket.SHUT_RDWR)
```
brainfunked: Class initialisation shouldn't do any action. It should just setup the object's initial state. Another method should be invoked to do the actual action. In this case, only the variables (including the socket object, but not the actual connection) should be initialised.

The actual writing to the socket should happen in a `write` method. This `write` method should open the socket connection, write the data and then close the connection.

tendrl/commons/message.py
```diff
@@ -0,0 +1,47 @@
...
+                self.caller = {"filename": (caller.filename.rsplit('/', 1))[1],
+                               "line_no": caller.lineno,
+                               "function": caller.function}
+            else:
+                self.caller = message["caller"]
+            if not self.validate():
+                self.timestamp =  datetime.datetime.now().isoformat()
+                self.priority = "debug"
+                self.publisher =  "node-agent"
+                # Invalid obj does not have node_id
```
brainfunked: Every message must have a node_id.

tendrl/commons/message.py
brainfunked: cluster_id is not mandatory. Jobs can occur outside a cluster context.

tendrl/commons/message.py
```diff
@@ -0,0 +1,47 @@
+import datetime
+from inspect import getframeinfo
+from inspect import stack
+import uuid
+
+
+class Message(object):
```
brainfunked: The Message class' interface should be as follows:

* Initialisation is done only when crafting a new message from scratch. This would be used by the entities that generate the messages. The initialise method should accept mandatory parameters directly, with perhaps a dict for the payload. Initialisation would thus automatically validate the presence of the mandatory parameters. This method should always generate a message_id itself.
* The class method `from_json` should be used when generating a message object from json. This would be the case in the node agent, when it reads a message from the socket. This method should raise an error clearly indicating that the json syntax is invalid, if applicable. This method must call the `validate` method (discussed below).
* The object should have a `to_json` method to dump the json format.
* The `validate` method should check for the existence of all the mandatory parameters, `message_id` among them. It must raise a validation error with the message indicating the precise problems.
* Whoever creates the message object using `from_json` would be responsible for catching the errors. In case of validation errors, it needs to initialise a new message object, whose payload would be the original invalid message object with the validation error messages. Eventually would be a good idea to create a way to properly handle nested objects, but this'll do for now.

brainfunked: Additionally, the `validate` method should check the attributes, like it currently does. It should not be validating json. Thus it must be called only after the attributes have been set.

GowthamShanmugam:
@brainfunked @shtripat @nthomas-redhat @anmolbabu @anmolsachan @r0h4n please review this, i have done all the changes.

tendrl/commons/event.py
```diff
@@ -0,0 +1,17 @@
...
+PATH = "/etc/tendrl/node_agent_socket"
+
+
+class Event(object):
+    def __init__(self, message):
+        self.sock = _socket.socket(_socket.AF_UNIX, _socket.SOCK_STREAM)
+        self.write(message)
+
+    def write(self, message):
+        self.sock.connect(PATH)
```
anmolbabu: what happens if there is a connection failure?

GowthamShanmugam: it will raise an exception, who is calling event  should take care of it

anmolbabu: how user gets to know?

For example, in case of alerting, the threshold detection happens @ collectd, the plugin invokes the Message framework below and then the framework's attempt to open a socket connection to node-agent exposed socket fails.
In this case the plugin gets to know about the exception and the maximum it can do is log the failure to collectd logs.But how can that be known by the user I am asking this more with this perspective, if its an error/exception happening internal to collectd, probably user going to collectd logs might seem logical but if its an issue very much specific to processes internal to tendrl, I would like there to be a way for user to be aware of it from the tendrl logging framework itself rather than from integration component logs.

Also consider another example, the ceph and gluster integrations will be responsible for detecting state changes of entities managed by them, consider if this connect fails, then what happens how can it be made known to user?

Or modules would terminate at the very init if the socket creation @ node-agent and connection @ other modules end failed is my assumption correct? In such cases also, how is it planned to log/notify error in such cases to user?

With socket path hard-coded and communication being local within the node itslef, I am not sure about likelihood of such a failure with packaging doing the necessary permissions to the sock file and ensuring existence of socket file.
But, I would like to take opinions from @Tendrl/tendrl-core here

GowthamShanmugam: @brainfunked please share your suggestion here

tendrl/commons/message.py
```diff
@@ -0,0 +1,87 @@
...
+    """
+    def __init__(self, priority, publisher, payload,
+                 message_id=None, timestamp=None, node_id=None,
+                 caller=None):
+        if message_id is None:
+            self.message_id = str(uuid.uuid4())
+            self.timestamp = datetime.datetime.now().isoformat()
+            # From which function, line and file error raised
+            caller = getframeinfo(stack()[1][0])
+            self.caller = {"filename": (caller.filename.rsplit('/', 1))[1],
```
r0h4n: Most Tendrl code runs inside a greenlet, how does getframeinfo work with that?

tendrl/commons/message.py
```diff
@@ -0,0 +1,87 @@
...
+
+    caller are None by default, when from_json function call
+    message old message_id, time_stamp and caller is populated
+    """
+    def __init__(self, priority, publisher, payload,
+                 message_id=None, timestamp=None, node_id=None,
+                 caller=None):
+        if message_id is None:
+            self.message_id = str(uuid.uuid4())
+            self.timestamp = datetime.datetime.now().isoformat()
```
r0h4n: Move this function to commons https://github.com/Tendrl/ceph-integration/blob/master/tendrl/ceph_integration/util.py#L7

use it for timestamp

tendrl/commons/message.py
```diff
@@ -0,0 +1,87 @@
...
+        # Check payload type is dict
+        if type(self.payload) != dict:
+            return False
+        # Check mandatory fields
+        if (self.priority not in Message.priorities.__dict__.values() or
+            self.publisher not in Message.publishers.__dict__.values() or
+            self.node_id is None or
+                "message" not in self.payload):
+            return False
+        if "job_id" in self.payload:
```
r0h4n: describe the validation process in a comment at start of function

tendrl/commons/message.py
```diff
@@ -0,0 +1,87 @@
...
+        return True
+
+
+# To serialize when json contains old message object
+def serialize_message(obj):
+    if isinstance(obj, Message):
+        serial = obj.__dict__
+        return serial
+    else:
+        raise TypeError("Type not serializable")
```
r0h4n: what type, please mention here

tendrl/commons/event.py
```diff
@@ -0,0 +1,25 @@
+import _socket
+from tendrl.commons.message import Message
+
+
+PATH = "/etc/tendrl/.node_agent_socket"
```
GowthamShanmugam: @r0h4n i will add this in config file

tendrl/commons/event.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class EventFailed(Exception):
+    pass
+
+
+class Event(object):
+    def __init__(self, message):
+        self.sock = _socket.socket(_socket.AF_UNIX, _socket.SOCK_STREAM)
+        self.write(message)
```
GowthamShanmugam: @r0h4n when you send the message to socket at that time it should be a json ... so in Event  __init__ you need to pass object, then write method will tack care converting obj into json_str using to_json() in Message class

tendrl/commons/message.py
```diff
@@ -0,0 +1,87 @@
...
+import uuid
+
+
+class Message(object):
+    # priority[0] is debug....priority[5] is critical
+    # Use priorities and publishers when call create message obj
+    priorities = type("priorities", (),
+                      {"DEBUG": "debug", "INFO": "info", "NOTICE": "notice",
+                       "WARNING": "warning", "ERROR": "error",
+                       "CRITICAL": "critical"})
```
GowthamShanmugam: @r0h4n it is works like enum, for example when you wants to a message you need to give like this

Message(Message.priorities.DEBUG, ....) so it will pass value "debug" to __init__()

GowthamShanmugam: @r0h4n these are fixed priorities, it is used to avoid spelling mistakes when we give priorities

tendrl/commons/message.py
GowthamShanmugam: @r0h4n it will tacking caller details from stack, How it will work means it will take last call of this function only. so it will give exactly who called this function

tendrl/commons/message.py
GowthamShanmugam: @r0h4n i will comment there

tendrl/commons/message.py
GowthamShanmugam: @r0h4n done

tendrl/commons/message.py
GowthamShanmugam: @r0h4n done

tendrl/commons/event.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class Event(object):
+    def __init__(self, message):
+        self.sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+        self.write(message)
+
+    def write(self, message):
+        try:
+            socket_path = tendrl_ns.config.data['socket_path']
```
r0h4n: change the name to "logging_socket_path"

tendrl/commons/event.py
```diff
@@ -0,0 +1,25 @@
...
+class EventFailed(Exception):
+    pass
+
+
+class Event(object):
+    def __init__(self, message):
+        self.sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+        self.write(message)
+
+    def write(self, message):
```
r0h4n: make this write private, _write

tendrl/commons/message.py
r0h4n: Where is it done? I dont see you using this for timestamp anywhere https://github.com/Tendrl/ceph-integration/blob/master/tendrl/ceph_integration/util.py#L7

tendrl/commons/event.py
GowthamShanmugam: @r0h4n done

tendrl/commons/event.py
GowthamShanmugam: @r0h4n Done

GowthamShanmugam:
@r0h4n ack i am doing that change changing timestamp

tendrl/commons/event.py
```diff
@@ -0,0 +1,25 @@
+from _socket import error as _socket_error
+from _socket import timeout as _socket_timeout
```
r0h4n: use gevent error and timeout, remove _socket

tendrl/commons/event.py
r0h4n: where?

tendrl/commons/event.py
r0h4n: still not done

requirements.txt
```diff
@@ -8,4 +8,5 @@ python-etcd
 python-dateutil
 PyYAML
 gevent>=1.0
-namespaces
\ No newline at end of file
+namespaces
+python-dateutil
```
GowthamShanmugam: @r0h4n ok

GowthamShanmugam:
@r0h4n all changes now i done correctly

tendrl/commons/message.py
```diff
@@ -0,0 +1,101 @@
...
+        return True
+
+
+# To serialize when json contains old message object
+def serialize_message(obj):
+    if isinstance(obj, Message):
+        serial = obj.__dict__
+        return serial
+    elif isinstance(obj, datetime.datetime):
+        serial = obj.isoformat()
```
r0h4n: timezone info maintained during serialization?

GowthamShanmugam: @r0h4n timezone can't serialize, so when passing object into socket  i have converted timezone into string, and then string again converted to datetime when jon is read from socket.

GowthamShanmugam: @r0h4n timezone can't serialize, so when passing object into socket  i have converted timezone into string, and then string again converted to datetime when jon is read from socket.

tendrl/commons/utils/tz_utils.py
```diff
@@ -0,0 +1,8 @@
+import datetime
```
r0h4n: rename this file "time_utils"

tendrl/commons/message.py
```diff
@@ -0,0 +1,101 @@
...
+        # Check payload type is dict
+        if type(self.payload) != dict:
+            return False
+        # Check mandatory fields
+        if (self.priority not in Message.priorities.__dict__.values() or
+            self.publisher not in Message.publishers.__dict__.values() or
+            self.node_id is None or
+                "message" not in self.payload):
+            return False
+        if "job_id" in self.payload:
```
r0h4n: "job_id": what is this field and give an example , is this the api job id?

flow_id: is this the uuid of the flow as given in the tendrl definitions?

anmolsachan: @r0h4n "job_id" is the field for the uuid of the Operation updates. For example this could be an update for any atom executed under a flow etc.
yes, "flow_id" is the  uuid of the flow as given in the tendrl definitions
Both are mandatory fields for the Operation updates

r0h4n: Ok, so the "job_id" is actually "request_id" which is present in the job payload and is generated by backend for all jobs which are not "new", it is not something the API generates.

Please change it to "request_id"

GowthamShanmugam:
@r0h4n changes are done

r0h4n:
@GowthamShanmugam Please send this PR against the "develop" branch, we are not merging in master directly

PR #145, opened 2017-01-27
Fixed issues related to flows and atoms initialization

shtripat:

### DIFF
--- a/tendrl/commons/jobs/__init__.py
+++ b/tendrl/commons/jobs/__init__.py
@@ -94,8 +94,8 @@ class JobConsumer(object):
     def invoke_flow(self, flow_fqn, job):
         # flow_fqn eg:tendrl.node_agent.objects.abc.flows.temp_flows
         if "tendrl" in flow_fqn and "objects" in flow_fqn:
-            obj_name, flow_name = flow_fqn.split(".objects.")[-1].split(
-                ".flows.")[-1].split(".")[-1]
+            obj_name = flow_fqn.split(".objects.")[-1].split(".")[0]
+            flow_name = flow_fqn.split(".flows.")[-1].split(".")[-1]
             flow = tendrl_ns.get_obj_flow(obj_name, flow_name)
             return flow(parameters=job['parameters'],
                         request_id=job['request_id']).run()
--- a/tendrl/commons/objects/atoms/__init__.py
+++ b/tendrl/commons/objects/atoms/__init__.py
@@ -19,11 +19,11 @@ class BaseAtom(object):
                                              self.obj.__name__)
         atom_def = obj_def.atoms[self.__class__.__name__]
 
-        self.inputs = inputs or atom_def['inputs']['mandatory']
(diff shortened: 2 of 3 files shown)


tendrl/commons/jobs/__init__.py
```diff
@@ -94,7 +94,9 @@ def stop(self):
     def invoke_flow(self, flow_fqn, job):
         # flow_fqn eg:tendrl.node_agent.objects.abc.flows.temp_flows
         if "tendrl" in flow_fqn and "objects" in flow_fqn:
-            obj_name, flow_name = flow_fqn.split(".objects.")[-1].split(
+            obj_name = flow_fqn.split(".objects.")[-1].split(
+                ".flows.")[-2]
```
r0h4n: the older code was doing the same, whats the difference?

shtripat: Earlier we were trying to get obj_name and flow_name both in the same call. The value of obj_name used to be correct but flow name was not. Flow name used toe be reported as e.g. `create_volume.CreateVolume` whereas we need only `CreateVolume`, so we get these values separately now. Thats all...

tendrl/commons/objects/atoms/__init__.py
```diff
@@ -19,11 +19,11 @@ def __init__(
                                              self.obj.__name__)
         atom_def = obj_def.atoms[self.__class__.__name__]
 
-        self.inputs = inputs or atom_def['inputs']['mandatory']
-        self.enabled = enabled or atom_def['enabled']
-        self.outputs = outputs or atom_def['outputs']
-        self.uuid = uuid or atom_def['uuid']
-        self.help = help or atom_def['help']
+        self.inputs = inputs or atom_def.get('inputs')['mandatory']
```
r0h4n: atom_def.get('inputs')['mandatory'] should be atom_def.get('inputs').get("mandatory")

shtripat: Done

tendrl/commons/jobs/__init__.py
```diff
@@ -94,7 +94,9 @@ def stop(self):
...
+            flow_name = flow_fqn.split(".objects.")[-1].split(
```
r0h4n: flow_name = flow_fqn.split(".flows.")[-1].split(".")[-1]

tendrl/commons/jobs/__init__.py
r0h4n: obj_name = flow_fqn.split(".objects.")[-1].split(".")[0]

much cleaner

shtripat: Done

tendrl/commons/jobs/__init__.py
shtripat: Done

shtripat:
Verified with latest create/delete pool changes and works as expected.

PR #148, opened 2017-01-31
Add null check for sds sync thread which can be optional

anmolbabu:


anmolbabu:

PR #150, opened 2017-02-01
Logging changes due to new logging framework 

anmolsachan:
tendrl-spec: Tendrl/specifications/pull/127 
reference-spec: Tendrl/specifications/pull/94


anmolsachan:
@r0h4n I will change the pr accordingly . I need your views on the following :
Though we can change the logging statements at every place but there is an issue with it. The new logging framework requires message_handler_thread to be started in ( /Tendrl/node-agent/pull/178/ tendrl/node_agent/manager/__init__.py ) . But before this there are a lot of imports from commons which contain log statements. What should be done about these, as these will fail because of unavailability of message handler ?

anmolsachan:
Sending to develop branch

PR #156, opened 2017-02-08
Logging implementation for tendrl

GowthamShanmugam:
tendrl-spec: logging_implementation

Signed-off-by: root <root@dhcp42-75.lab.eng.blr.redhat.com>


GowthamShanmugam:
@nthomas-redhat @anmolbabu @brainfunked @anmolsachan @shtripat  please review this

tendrl/commons/event.py
```diff
@@ -0,0 +1,25 @@
...
+    def _write(self, message):
+        try:
+            socket_path = tendrl_ns.config.data['logging_socket_path']
+            self.sock.connect(socket_path)
+            json_str = Message.to_json(message)
+            self.sock.send(json_str)
+            self.sock.close()
+        except (socket_error, socket_timeout,
+                TypeError) as ex:
+            raise EventFailed(ex)
```
anmolsachan: Here we are raising the event if the socket fails. How should this situation be handled ? Should the caller of the function should handle this every time he logs ? @nthomas-redhat @shtripat please review

tendrl/commons/message.py
```diff
@@ -0,0 +1,102 @@
...
+                      {"DEBUG": "debug", "INFO": "info", "NOTICE": "notice",
+                       "WARNING": "warning", "ERROR": "error",
+                       "CRITICAL": "critical"})
+    publishers = type("publishers", (),
+                      {"CEPH_INTEGRATION": "ceph-integration",
+                       "GLUSTER_INTEGRATION": "gluster-integration",
+                       "NODE_AGENT": "node-agent",
+                       "COMMONS": "commons",
+                       "ALERTING": "alerting",
+                       "PERFORMANCE_MONITORING": "performance-monitoring"})
```
brainfunked: Commons must not refer to any other components. The components creating the object from the Message class should inject their own publisher id. Each component should have it's own hardcoded publisher id.

tendrl/commons/message.py
```diff
@@ -0,0 +1,102 @@
...
+import uuid
+
+
+class Message(object):
+    # priority[0] is debug....priority[5] is critical
+    # Use priorities and publishers when call create message obj
+    priorities = type("priorities", (),
+                      {"DEBUG": "debug", "INFO": "info", "NOTICE": "notice",
+                       "WARNING": "warning", "ERROR": "error",
+                       "CRITICAL": "critical"})
```
brainfunked: What is the point of mapping the same string in different cases?

tendrl/commons/message.py
```diff
@@ -0,0 +1,102 @@
...
+        as job_updates, job updates should have flow id,
+        Check priorities and publishers are correct
+        """
+
+        # Check payload type is dict
+        if type(self.payload) != dict:
+            return False
+        # Check mandatory fields
+        if (self.priority not in Message.priorities.__dict__.values() or
+            self.publisher not in Message.publishers.__dict__.values() or
```
brainfunked: Publisher validation doesn't seem necessary to me. If you really want to do it, it must not be based on the hardcoded list of publishers in commons. That defeats the entire purpose of keeping commons as a generic library. In any case, if you still really want to do it, the publishers will have to put their id somewhere in a global listing in etcd and you'll need to validate from there. I don't think it's worth the effort for now.

tendrl/commons/message.py
```diff
@@ -0,0 +1,102 @@
...
+            self.node_id is None or
+                "message" not in self.payload):
+            return False
+        if "request_id" in self.payload:
+            if "flow_id" not in self.payload:
```
nthomas-redhat: How you ensure everyone adhere to the right format of payload?
Or do we need to write separate 'payload' class add it as composition in Message class?
@r0h4n @brainfunked

tendrl/commons/event.py
```diff
@@ -0,0 +1,25 @@
+from gevent import socket
+from gevent.socket import error as socket_error
+from gevent.socket import timeout as socket_timeout
+from tendrl.commons.message import Message
+
+
+class EventFailed(Exception):
+    pass
```
nthomas-redhat: Why 'pass' ? you need to have a proper message send out here

tendrl/commons/event.py
```diff
@@ -0,0 +1,25 @@
...
+        self.sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+        self._write(message)
+
+    def _write(self, message):
+        try:
+            socket_path = tendrl_ns.config.data['logging_socket_path']
+            self.sock.connect(socket_path)
+            json_str = Message.to_json(message)
+            self.sock.send(json_str)
+            self.sock.close()
```
nthomas-redhat: This needs to be added in a 'finally' block so that it will be taken care when exception occurs

tendrl/commons/message.py
GowthamShanmugam: @brainfunked It is like enum, they can pass like
     Message(Message.priorities.INFO, Message.publishres.COMMOM, {"message": "some message"})

so   Message.priorities.INFO will give info to message parameter, it is useful for avoid spell mistake and only this function can use.

tendrl/commons/event.py
```diff
@@ -0,0 +1,25 @@
...
+
+class Event(object):
+    def __init__(self, message):
+        self.sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+        self._write(message)
+
+    def _write(self, message):
+        try:
+            socket_path = tendrl_ns.config.data['logging_socket_path']
+            self.sock.connect(socket_path)
```
nthomas-redhat: Is this a good idea to create/connect/delete sockets for every log message? I am just thinking in terms of volume of these operations. Or is it a good idea to think about some kind of socket client pooling?
@brainfunked @r0h4n

tendrl/commons/event.py
nthomas-redhat: @anmolsachan @GowthamShanmugam , what are scenarios where this can fail?

tendrl/commons/message.py
GowthamShanmugam: @brainfunked i will change it

tendrl/commons/message.py
GowthamShanmugam: @brainfunked i will remove publisher validation

tendrl/commons/message.py
```diff
@@ -0,0 +1,102 @@
...
+    # Use priorities and publishers when call create message obj
+    priorities = type("priorities", (),
+                      {"DEBUG": "debug", "INFO": "info", "NOTICE": "notice",
+                       "WARNING": "warning", "ERROR": "error",
+                       "CRITICAL": "critical"})
+    publishers = type("publishers", (),
+                      {"CEPH_INTEGRATION": "ceph-integration",
+                       "GLUSTER_INTEGRATION": "gluster-integration",
+                       "NODE_AGENT": "node-agent",
+                       "COMMONS": "commons",
```
GowthamShanmugam: @nthomas-redhat i will change it

tendrl/commons/message.py
```diff
@@ -0,0 +1,102 @@
...
+        self.node_id = tendrl_ns.node_context.node_id
+        self.payload = payload
+
+    @staticmethod
+    def from_json(json_str):
+        message_json = json.loads(json_str)
+        message = Message(**message_json)
+        if not message.validate():
+            # Invalid message logged as debug
+            message_new = Message(Message.priorities.DEBUG,
```
GowthamShanmugam: @nthomas-redhat message_id and called details are automatically taken by message class. After message_json_str is received by node_agent socket handler, the message_json_str is passed to from_json function in message to conver json_str to message object, at that time caller and message_id is not created again, old caller and mesage id is passed and assigned.

tendrl/commons/message.py
GowthamShanmugam: @nthomas-redhat here the idea is payload must have message, request id is optional, but if request id came then flow_id should come.
request is not mandatory. message is mandatory. you can give anything in message part.
for example:
     @anmolbabu wants to pass alert details using message object, so what he can do is all the alert details he can pass under message part in payload.
    for job updates payload message part contains job update details and give request_id and flow_id. so if request_id is present the message handler automatically understand it is job updates.

GowthamShanmugam: @nthomas-redhat only for job updated there is some extra fields in payload, otherwise for all everything is comes under message part in payload

tendrl/commons/message.py
```diff
@@ -0,0 +1,102 @@
...
+
+        return True
+
+
+# To serialize when json contains old message object
+def serialize_message(obj):
+    if isinstance(obj, Message):
+        serial = obj.__dict__
+        return serial
+    elif isinstance(obj, datetime.datetime):
```
GowthamShanmugam: @nthomas-redhat it is comes under def serialize_message(obj) method it is not in message class if is different function. Use of this method is we cant pass message object into socket directly, we can to convert message object into json_str. Problem is when convert message object into json_str it gives serialization error.
problem is datetime and some times old object is stored in new object in payload message part, so these to can't serialize.
During serialize this method is called automatically and convert those two into serialize

GowthamShanmugam: @nthomas-redhat 
 def to_json(message):
         return json.dumps(message.__dict__, default=serialize_message)
Here it is used

tendrl/commons/event.py
GowthamShanmugam: @nthomas-redhat i will change it

tendrl/commons/event.py
GowthamShanmugam: @nthomas-redhat socket is created only once, which is return in node agent, here event class will connect with socket which is running already and pass send the message into it

tendrl/commons/event.py
GowthamShanmugam: @nthomas-redhat i wil change it

GowthamShanmugam:
@nthomas-redhat if socket file is deleted then event class can't connect with socket, at that time exception came. If socket creation is not happened in node_agent then event cant cant connect with socket at that time exception will come.

tendrl/commons/event.py
r0h4n: @nthomas-redhat is correct here, please evaluate below socket connection pool libraries or make your own socket connection pool 

- https://pypi.python.org/pypi/socketpool
- https://pypi.python.org/pypi/geventconnpool/0.2.1

tendrl/commons/message.py
r0h4n: I agree with @nthomas-redhat , need a payload class in commons

tendrl/commons/event.py
GowthamShanmugam: @nthomas-redhat @r0h4n unix no implementations for unix socket connection pool, i have discussed with @brainfunked he said unix socket connection are very light weight so no need connection pool for that, and i have searched all pool connection are only for tcp connection not for unix socket connection

GowthamShanmugam:
@nthomas-redhat @brainfunked @shtripat @nnDarshan @anmolbabu @r0h4n @TimothyAsir @anmolsachan  please review

GowthamShanmugam:
@nthomas-redhat @shtripat @nnDarshan @r0h4n @brainfunked @anmolbabu @TimothyAsir please review

tendrl/commons/manager/__init__.py
```diff
@@ -14,27 +14,35 @@ def __init__(
...
             central_store_thread,
+            message_handler_thread=None,
     ):
         self._central_store_thread = central_store_thread
         self._sds_sync_thread = sds_sync_thread
         self._job_consumer_thread = jobs.JobConsumerThread()
+        self._message_handler_thread = message_handler_thread
 
     def stop(self):
         LOG.info("%s stopping" % self.__class__.__name__)
```
GowthamShanmugam: @brainfunked  i will change it

tendrl/commons/manager/__init__.py
```diff
@@ -14,27 +14,35 @@ def __init__(
...
         LOG.info("%s starting" % self.__class__.__name__)
+        if self._message_handler_thread is not None:
+            self._message_handler_thread.start()
         self._central_store_thread.start()
         if self._sds_sync_thread is not None:
             self._sds_sync_thread.start()
         self._job_consumer_thread.start()
 
     def join(self):
         LOG.info("%s joining" % self.__class__.__name__)
```
GowthamShanmugam: @brainfunked i will change it

tendrl/commons/manager/__init__.py
GowthamShanmugam: @brainfunked This log message is removed in @anmolsachan patch because it is happening before node socket start

PR #162, opened 2017-02-17
Removed old style job status update

shtripat:
Specific module's atoms and flows would update the job status
properly using the logging framework in Tendrl now.


shtripat:

PR #164, opened 2017-02-18
Logging fix to accomdate external module integration

anmolbabu:
Logging fix to accomdate external module integration(collectd)


anmolbabu:
@Tendrl/tendrl-core @GowthamShanmugam This patch is required for external modules like the collectd plugin to post threshold breach info to tendrl. The collectd plugin has no tendrl_ns object and hence will pass the accessed tendrl_ns attributes as parameters here. Please review...

GowthamShanmugam:
LGTM

anmolbabu:

tendrl/commons/message.py
```diff
@@ -3,8 +3,14 @@
 from inspect import getframeinfo
 from inspect import stack
 import json
+import sys
 from tendrl.commons.utils.time_utils import now
+sys.path.remove('/usr/lib64/collectd')
 import uuid
+sys.path.append('/usr/lib64/collectd')
```
r0h4n: Are you sure collectd python modules are only found in "/usr/lib64/collectd"  and not at places like "/usr/lib/collectd" or any random places?

anmolbabu: Collectd's plugins are found only in that path

PR #171, opened 2017-02-23
check if python path exists before removing it

anmolbabu:

### DIFF
--- a/tendrl/commons/message.py
+++ b/tendrl/commons/message.py
@@ -5,9 +5,13 @@ from inspect import stack
 import json
 import sys
 from tendrl.commons.utils.time_utils import now
-sys.path.remove('/usr/lib64/collectd')
+is_collectd_imported = False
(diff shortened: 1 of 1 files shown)


anmolbabu:
@r0h4n Please review and merge if it looks good

tendrl/commons/message.py
```diff
@@ -5,7 +5,8 @@
 import json
 import sys
 from tendrl.commons.utils.time_utils import now
-sys.path.remove('/usr/lib64/collectd')
+if '/usr/lib64/collectd' in sys.path:
+    sys.path.remove('/usr/lib64/collectd')
 import uuid
 sys.path.append('/usr/lib64/collectd')
```
r0h4n: The append should only happen if remove was called

anmolbabu:
@r0h4n Incorporated your comment. Please review

PR #172, opened 2017-02-26
Added check for AtomExecutionFailedError to fail the task

shtripat:
Also added check if mandatory arguments are passed for a
flow while execution.

tendrl-bug-id: Tendrl/api#33


shtripat:

PR #176, opened 2017-03-02
SSH-key configuration utilities

GowthamShanmugam:
tendrl-bug-id: Tendrl/specifications#141
tendrl-spec: ssh_configuration


GowthamShanmugam:
For this psutil (pip package) and  libselinux-python (yum package) is required, but for rhel names are different python-psutil  and  libselinux i am going to raise bug for this. @TimothyAsir @shtripat @r0h4n @nnDarshan @brainfunked

GowthamShanmugam:
@TimothyAsir @shtripat @nnDarshan @r0h4n @brainfunked please review this

tendrl/commons/utils/SSHUtil.py
```diff
@@ -0,0 +1,43 @@
+import psutil
+from tendrl.commons.event import Event
+from tendrl.commons.message import Message
+from tendrl.commons.utils import cmd_utils
+
+
+def sshd_status():
+    cmd = cmd_utils.Command("cat /var/run/sshd.pid")
```
brainfunked: pid file is not guaranteed to exist. Furthermore, locations change. Since we're relying on systemd, a much more reliable method would be to use `systemctl show sshd.service` and grab the value of MainPID.

tendrl/commons/utils/generate_ssh.py
```diff
@@ -0,0 +1,47 @@
+from ansible_module_runner import AnsibleExecutableGenerationFailed
+from ansible_module_runner import AnsibleRunner
+from tendrl.commons.event import Event
+from tendrl.commons.message import Message
+
+ANSIBLE_MODULE_PATH = "core/system/user.py"
```
brainfunked: Does this module guarantee that the SSH key will be generated if the user already exists?

tendrl/commons/utils/SSHUtil.py
GowthamShanmugam: @brainfunked problem is we can't use pip symbol in ansible command module, so grep is not possible

tendrl/commons/utils/generate_ssh.py
GowthamShanmugam: @brainfunked yes i tried, if the user is exist it will create only ssh-key, if the user is not exist it will create user with ssh-key

GowthamShanmugam:
@brainfunked Changes done

tendrl/commons/utils/ssh_util.py
```diff
@@ -0,0 +1,64 @@
...
+        Event(
+            Message(
+                priority="error",
+                publisher="commons",
+                payload={"message": err}
+            )
+        )
+        return None, err
+
+def find_pid(out):
```
brainfunked: This method should be private and not accessible to anyone using the util.

tendrl/commons/utils/ssh_util.py
```diff
@@ -0,0 +1,64 @@
...
+    if not err:
+        sshd = {}
+        out = out.split("\n")
+        pid = find_pid(out)
+        if pid == 0:
+            Event(
+                Message(
+                    priority="error",
+                    publisher="commons",
+                    payload={"message": "sshd service is not run"}
```
brainfunked: "is not running"

tendrl/commons/utils/ssh_util.py
```diff
@@ -0,0 +1,64 @@
...
+                )
+            )
+            return sshd, "sshd service is not run"
```
brainfunked: No returns in between. Write pre-determined status values in a constant and use one of them in the dict you return at the end of the method. You can short circuit in between if the checks further down don't make any sense, but return must always be a dict with the same format each time.

tendrl/commons/utils/ssh_util.py
```diff
@@ -0,0 +1,64 @@
...
+
+        [pconn(fd=3, family=10, type=1, laddr=('0.0.0.0', 22),
+        raddr=('::1', 54960), status='LISTEN')]
+        '''
+        result = [con for con in p.connections() if con.status ==
+                  psutil.CONN_LISTEN and con.laddr[0] == "0.0.0.0"]
+        if result == []:
+            Event(
+                Message(
+                    priority="error",
```
brainfunked: warning, not error.

GowthamShanmugam:
@brainfunked done all changes

tendrl/commons/utils/ssh_util.py
```diff
@@ -0,0 +1,64 @@
+import psutil
+from tendrl.commons.event import Event
+from tendrl.commons.message import Message
+from tendrl.commons.utils import cmd_utils
+
+
+def sshd_status():
```
brainfunked: This util must always return a specific data structure, which would seem to be a dict with keys "name", "port" and "status". This must always be the return value. If there are any errors, the values should probably be Null or empty, whatever is the appropriate python convention. In none of the error conditions are you supposed to return directly with just a string. Use log messages to indicate errors.

tendrl/commons/utils/ssh_util.py
```diff
@@ -0,0 +1,64 @@
...
+    out, err, rc = cmd.run(tendrl_ns.config.data[
+                           'tendrl_ansible_exec_file'])
+    if not err:
+        sshd = {}
+        out = out.split("\n")
+        pid = find_pid(out)
+        if pid == 0:
+            Event(
+                Message(
+                    priority="error",
```
brainfunked: warning

tendrl/commons/utils/ssh_util.py
```diff
@@ -0,0 +1,64 @@
...
+                  psutil.CONN_LISTEN and con.laddr[0] == "0.0.0.0"]
+        if result == []:
+            Event(
+                Message(
+                    priority="error",
+                    publisher="commons",
+                    payload={"message": "Unable to find port number"}
+                )
+            )
+            return sshd, "Unable to find port number"
```
brainfunked: Again, no return in between.

tendrl/commons/utils/ssh/generate_key.py
```diff
@@ -0,0 +1,47 @@
+from ansible_module_runner import AnsibleExecutableGenerationFailed
+from ansible_module_runner import AnsibleRunner
+from tendrl.commons.event import Event
+from tendrl.commons.message import Message
+
+ANSIBLE_MODULE_PATH = "core/system/user.py"
+
+
+class GenerateSSH(object):
```
brainfunked: Update the class names to match the file names.

tendrl/commons/utils/ssh/authorize_key.py
```diff
@@ -0,0 +1,47 @@
+from ansible_module_runner import AnsibleExecutableGenerationFailed
+from ansible_module_runner import AnsibleRunner
+from tendrl.commons.event import Event
+from tendrl.commons.message import Message
+
+ANSIBLE_MODULE_PATH = "core/system/authorized_key.py"
+
+
+class AuthorizeKey(object):
```
brainfunked: Determine the output data structure for this utility. Make sure that all return values use the same data structure. Write error messages into logs at warning priority.

tendrl/commons/utils/ssh/generate_key.py
GowthamShanmugam: @brainfunked ok

tendrl/commons/utils/ssh_util.py
GowthamShanmugam: @brainfunked i will change as per your suggestion

tendrl/commons/utils/ssh_util.py
GowthamShanmugam: @brainfunked done

tendrl/commons/utils/ssh_util.py
GowthamShanmugam: @brainfunked done

tendrl/commons/utils/ssh_util.py
GowthamShanmugam: @brainfunked ok

tendrl/commons/utils/ssh_util.py
GowthamShanmugam: @brainfunked ok

tendrl/commons/utils/ssh_util.py
GowthamShanmugam: @brainfunked ok

tendrl/commons/utils/ssh_util.py
GowthamShanmugam: @brainfunked ok

tendrl/commons/utils/ssh/authorize_key.py
GowthamShanmugam: @brainfunked ok

GowthamShanmugam:
@shtripat @brainfunked @r0h4n @nnDarshan @nthomas-redhat  please review
Authorize_key module works when selinux is  disabled

PR #184, opened 2017-03-07
Psutil added as a dependency for ssh_utils

GowthamShanmugam:


GowthamShanmugam:
@shtripat @r0h4n

GowthamShanmugam:
@TimothyAsir please review

GowthamShanmugam:
I have verified this package 
for yum in centos and rhel is python-psutil
for pip psutil

Issue #177, opened 2017-03-02
PEP8 check fails


mbukatov:
[PEP8 code style](https://www.python.org/dev/peps/pep-0008/) is not followed which makes `pep8` check fail during unit test execution, see [latest report from Travis CI](https://travis-ci.org/Tendrl/commons/jobs/206963318).

There are few serious issues among errors, such as:

```
tendrl/commons/message.py:50:28: F821 undefined name 'tendrl_ns'
            self.node_id = tendrl_ns.node_context.node_id
```

Note: to run pep8 check yourself locally, run `tox -e pep8` in current master branch, the environment and the check would be the same as run by Travis CI.

r0h4n:
Fixed: https://github.com/Tendrl/commons/commit/9a8fe461af3c3b8d9afaecce0f00469bab8c733d

PR #200, opened 2017-03-08
fix most errors reported by flake8

mbukatov:
All issues reported by flake8 has been addressed with two exceptions:

* 28 occurrences of error F821 undefined name 'tendrl_ns'
* one F841 local variable 'ex' is assigned to but never used

Related to https://github.com/Tendrl/commons/issues/177


mbukatov:
@r0h4n I checked the diff against develop and rebase is impractical (it would basically mean to do the work over again), but I could try to change this branch to be mergeable into develop branch (basically checking merge conflicts and reverting changes which leads to conflicts) - so that there would be some unfixed pep8 issues left to unblock clean merge into develop branch

mbukatov:
@r0h4n Let's do the fix this way: ping me when you are ready for these fixes to be merged, then I will revert all changes which conflicts with develop branch into a single commit so that you would be able to merge this into develop cleanly, and then just after that, you fix all the remaining pep8 issues so that you will be able to start enforcing pep8 check immediately (not merging any commits which don't pass the pep8 check).

Any other solution just doesn't work and would require those fixes to be done again.

Sooner we do this, the better.

PR #207, opened 2017-03-09
Updated to handle "tendrl" name space flows

nnDarshan:
with this PR current_ns will be correctly populated with NS even when namespace is "tendrl".


nnDarshan:

nnDarshan:
with this PR the run: value in import cluster job would  be "tendrl.flows.ImportCluster"

r0h4n:
I have fixed this in latest commons:develop

PR #225, opened 2017-03-10
Corrected object specific flow loading logic

shtripat:

### DIFF
--- a/tendrl/commons/flows/__init__.py
+++ b/tendrl/commons/flows/__init__.py
@@ -23,9 +23,9 @@ class BaseFlow(object):
         cls_name = self.__class__.__name__
         if hasattr(self, "obj"):
             obj_name = self.obj.__name__
-            self._defs = self._ns.get_obj_flow_definition(obj_name,
+            self._defs = self.obj._ns.get_obj_flow_definition(obj_name,
                                                               cls_name)
-            self.to_str = "%s.objects.%s.flows.%s" % (self._ns.ns_name,
+            self.to_str = "%s.objects.%s.flows.%s" % (self.obj._ns.ns_name,
                                                       obj_name,
                                                       cls_name)
         else:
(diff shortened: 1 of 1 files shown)


shtripat:

tendrl/commons/flows/__init__.py
```diff
@@ -115,7 +115,10 @@ def _execute_atom(self, atom_fqdn):
             ns, atom_name = atom_fqdn.split(".atoms.")
             ns, obj_name = ns.split(".objects.")
 
-            runnable_atom = self._ns.get_atom(obj_name, atom_name)
+            if hasattr(self, "obj"):
+                runnable_atom = self.obj._ns.get_atom(obj_name, atom_name)
+            else:
+                runnable_atom = self._ns.get_atom(obj_name, atom_name)
```
r0h4n: Its the same NS, why do you require it to be called as self.obj._ns?

shtripat: In case of object specific atoms like `object.Pool.atoms.delete.Delete, it was showing error `"self._ns" ` not defined. By debugging I found that `_ns` is defined under `self.obj` and so the changes.
Without this I was not able to test delete/resize/update pools and rbds.

nnDarshan: Even the gluster integration fix patch: https://github.com/Tendrl/gluster-integration/pull/165 was tested including this PR. Without this it was not working

shtripat: Thanks Darshan for confirmtaion.

r0h4n: "object._ns" : https://github.com/Tendrl/commons/blob/develop/tendrl/commons/__init__.py#L88
"atom._ns" : https://github.com/Tendrl/commons/blob/develop/tendrl/commons/__init__.py#L113
"flow._ns": https://github.com/Tendrl/commons/blob/develop/tendrl/commons/__init__.py#L124

@shtripat @nnDarshan could you guys re-check if this is still required, We do have the _ns object available for objects, atoms and flows

r0h4n:
Fixed: https://github.com/Tendrl/commons/commit/88b5188314b7f55670ac0c2f2df623205e58f468

PR #226, opened 2017-03-10
Corrected the config file creation while import cluster flow

shtripat:
The integrations specific config files were having wrong data
for etcd_port and etcd_connection. Also the attribute
logging_socket_path was missing. Added the same.


shtripat:

PR #169, opened 2017-02-21
Added exception message class for handle exceptions

GowthamShanmugam:
tendrl-bug-id: Tendrl/commons#168
tendrl-spec: logging_implementation

Signed-off-by: GowthamShanmugam <gshanmug@redhat.com>

PR #259, opened 2017-03-17
Exception message class for handle exception

GowthamShanmugam:
tendrl-bug-id: Tendrl/commons#169


GowthamShanmugam:
@r0h4n @shtripat review please

PR #262, opened 2017-03-19
Commons fixes

anmolbabu:
This PR fixes the following
1. NodeContext is used in job framework to detect tags operated on by tendrl's integration to determine 
    if the new job should be executed by the integration. And write of NodeContext is exclusive only to
    node-agent. In such a case, if integrations of tendrl like node-monitoring which has no rights to save  
    node-context to etcd can't execute a job. So, since the node_context containing tags is in-memory,
    instead of loading the NodeContext from etcd, the in-memory node-context can be used.
2. ExceptionMessage class used to yaml.safe_dump an exception object but that fails so changed it to 
    yaml.safe_dump excetion object's message attribute.
3. NodeContext always assumes tags to be present which might not be the case for modules like 
    performance-monitoring or alerting. So, changed the way the key was accessed.


anmolbabu:
@shtripat @nthomas-redhat @r0h4n @GowthamShanmugam Please review

tendrl/commons/jobs/__init__.py
```diff
@@ -51,9 +51,7 @@ def _run(self):
 
                         # Job routing
                         if raw_job.get("payload", {}).get("tags", []):
-                            NS.node_context = NS.node_context.load()
-                            tags = json.loads(NS.node_context.tags)
-                            if set(tags).isdisjoint(
```
r0h4n: You'll need to load() the node context because service tags are added during node sync in addition to tags mentioned in config. This change is not required

tendrl/commons/objects/node_context/__init__.py
```diff
@@ -24,7 +24,7 @@ def __init__(self, machine_id=None, node_id=None, fqdn=None,
         self.machine_id = machine_id or self._get_machine_id()
         self.node_id = node_id or self._get_node_id() or self._create_node_id()
         self.fqdn = fqdn or socket.getfqdn()
-        self.tags = tags or NS.config.data['tags']
+        self.tags = tags or NS.config.data.get('tags') or []
```
r0h4n: It is mandatory to define a default tag for each tendrl backend component. This change is not required

tendrl/commons/jobs/__init__.py
anmolbabu: write of NodeContext is exclusive only to
node-agent. In such a case, if integrations of tendrl like node-monitoring which has no rights to save
node-context to etcd can't execute a job. So, since the node_context containing tags is in-memory,
instead of loading the NodeContext from etcd, I thought the in-memory node-context can be used.

anmolbabu: How would you suggest, would be the ideal fix for this?

tendrl/commons/objects/node_context/__init__.py
anmolbabu: Ok but, I felt the tags are required when executing jobs but the problem is for those which don't themselves execute any jobs performance-monitoring for example.. Without, this change we will then need dummy tag or None equivalent tag.. Please suggest

tendrl/commons/jobs/__init__.py
r0h4n: NodeContext object is in commons, it can be accessed by any backend component at NS.node_context. 

Is the current code causing any traceback? This change is not required

tendrl/commons/objects/node_context/__init__.py
r0h4n: You'll need to use the tag tendrl/performance_monitoring and tendrl/node_monitoring and similar for alerting

tendrl/commons/jobs/__init__.py
anmolbabu: There's no exception just that the node-monitoring will not execute any job. Because, NodeContext is loaded from a path in etcd that is not written to by the node-monitoring and hence node-monitoring corresponding tags won't be there in that path and hence when loaded from etcd this tag won't be present and hence node-monitoring never executes any job targeted for it with specific tag "tendrl/monitoring" or "tendrl/node-monitoring" (and instead only executes those jobs that are meant for node-agent)

tendrl/commons/message.py
```diff
@@ -153,7 +153,7 @@ def serialize_message(obj):
         serial = obj.isoformat()
         return serial
     elif isinstance(obj, Exception):
-        return yaml.safe_dump(obj)
+        return yaml.safe_dump(obj.message)
```
GowthamShanmugam: Actually what here is happening is we are passing exception object and for serialization we are taking exception message only. I feel we have to keep entire exception object, that why i gave yaml.safe_dump(obj). what is your opinion @anmolbabu

anmolbabu: No yaml.safe_dump of exception object fails.

>>> ex = Exception('Exception caught')
>>> from ruamel import yaml
>>> yaml.safe_dump(ex)
  File "<stdin>", line 1, in <module>
  File "/usr/lib64/python2.7/site-packages/ruamel/yaml/main.py", line 278, in safe_dump
    return dump_all([data], stream, Dumper=SafeDumper, **kwds)
  File "/usr/lib64/python2.7/site-packages/ruamel/yaml/main.py", line 230, in dump_all
    dumper.represent(data)
  File "/usr/lib64/python2.7/site-packages/ruamel/yaml/representer.py", line 45, in represent
    node = self.represent_data(data)
  File "/usr/lib64/python2.7/site-packages/ruamel/yaml/representer.py", line 89, in represent_data
    node = self.yaml_representers[None](self, data)
  File "/usr/lib64/python2.7/site-packages/ruamel/yaml/representer.py", line 325, in represent_undefined
    raise RepresenterError("cannot represent an object: %s" % data)
ruamel.yaml.representer.RepresenterError: cannot represent an object: Exception caught
>>> yaml.safe_dump(ex.message)
'Exception caught\n...\n'

GowthamShanmugam: may i know how you passed exception in payload, can you please share that line
you have to give like this for example:
try:
    raise ValueError("Some error")  # i am creating value error manually
except ValueError as ex:  # here ex is important (here ex is exception object not string)
    Event(
        ExceptionMessage(
            priority="info" (any priority)
            publisher="" (some publisher name)
            payload={"message": "some exception related message", exception: ex}
        )
    )

anmolbabu: You can view it @ https://github.com/Tendrl/alerting/pull/41/files#diff-832c56a6c92c3a3783467abf5cef02c1R105

anmolbabu: @GowthamShanmugam Also it fails with the error "ruamel.yaml.representer.RepresenterError: cannot represent an object: Exception caught" as in my previous comment. That's why I created a python exception object and tried to yaml.safe_dump the exception object and it fails with same error as in my previous comment.

GowthamShanmugam: @anmolbabu can please try with yaml.dump(obj) instead of safe_dump

r0h4n:
Please remove the node context changes from this PR

tendrl/commons/objects/node_context/__init__.py
anmolbabu: Done

tendrl/commons/message.py
anmolbabu: @GowthamShanmugam I tried it and it seems to work but it gives unrelated python lingos which might not be desirable. See below
>>> from ruamel import yaml
>>> ex = Exception('Error!!')
>>> yaml.dump(ex)
'!!python/object/apply:exceptions.Exception [Error!!]\n'

anmolbabu: I am changing it to dump of exception object itself. But  @r0h4n @shtripat  @nthomas-redhat  ^^ What do you suggest

anmolbabu:
@r0h4n Done

tendrl/commons/message.py
r0h4n: The problem here is that ruamel.yaml or pyyaml wont understand how to represent your custom Exception object. Check below two fixes and make similar fixes here, youll need to add a way to represent your Exception object and let the yaml.dump know about that

- https://github.com/openshift/openshift-ansible/pull/3156/files
- https://github.com/ansible/ansible/blob/devel/lib/ansible/parsing/yaml/dumper.py#L31

PR #272, opened 2017-03-21
Removed logging of KeyNotFound error

shtripat:
This is not required as initially when there is no cluster
synced/created, these errors would keep logging every few
seconds and this would hog the log.


shtripat:

PR #273, opened 2017-03-21
Modified authorized_key util for work without disabling selinux

GowthamShanmugam:
tendrl-bug-id: Tendrl/commons#274


GowthamShanmugam:
@r0h4n @shtripat

tendrl/commons/utils/ssh/authorize_key.py
```diff
@@ -57,7 +72,7 @@ def run(self):
                                             str(e.message))}
                 )
             )
-        if err is not "":
+        if err != "" and err != "ValueError('No JSON object could be decoded',)":
```
GowthamShanmugam: Because of selinux it give this error but key is copied successfullt

r0h4n:
File an actual issue and explain the issue

GowthamShanmugam:
@r0h4n ok

tendrl/commons/utils/ssh/authorize_key.py
r0h4n: What is happening here?

tendrl/commons/utils/ssh/authorize_key.py
```diff
@@ -34,19 +35,33 @@ def run(self):
         output:
             True/False, error
         """
+        if self.attributes["user"] is 'root':
+            try:
+                os.mkdir("/root/.ssh", 700)
+            except OSError:
+                pass
+            open('/root/.ssh/authorized_keys', 'a').close()
```
GowthamShanmugam: @r0h4n Because if authorized_keys file is not present then first run it is creating authorized_keys file, but key is not copied, next time run it copying key inside authorize_key file, so i am creating authorize_key file if it is not present, so in first run copy key will done successfully

GowthamShanmugam:
@r0h4n if selinux is enabled then first time it gives error like this but key is copied successfully, so i am skipping this error.

nthomas-redhat:
> @r0h4n if selinux is enabled then first time it gives error like this but key is copied successfully, so i am skipping this error "ValueError('No JSON object could be decoded',)".
> 
> this solution is for temporary, we have to find out why ansible gives error when selinux is enable, then we can change.
Does this work if the selinux in 'permissive' mode?

GowthamShanmugam:
@nthomas-redhat not working in permissive mode also, it is working only when selinux is completely disabled

r0h4n:
I am still doubtful of this issue, can you please provide the ansible version you are seeing this issue for?

GowthamShanmugam:
@r0h4n i am using ansible 2.2.1.0

r0h4n:
And what platform? Give us entire details

PR #275, opened 2017-03-22
TTL on the message directories

GowthamShanmugam:

### DIFF
--- a/tendrl/commons/etcdobj/__init__.py
+++ b/tendrl/commons/etcdobj/__init__.py
@@ -70,7 +70,7 @@ class _Server(object):
 
         self.client = client
 
-    def save(self, obj):
+    def save(self, obj, ttl=None):
         """Save an object.
 
         :param obj: An instance that subclasses EtcdObj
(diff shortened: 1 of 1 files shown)


GowthamShanmugam:
@r0h4n i have tried to add TTL as attribute in ETCDOBJ but other child object classes gives exception like

GowthamShanmugam:

tendrl/commons/etcdobj/__init__.py
```diff
@@ -70,15 +70,18 @@ def _verify_client(self, client):
...
 
         :param obj: An instance that subclasses EtcdObj
         :type obj: EtcdObj
         :returns: The same instance
         :rtype: EtcdObj
         """
-        for item in obj.render():
+        render = obj.render()
+        if ttl != None:
+            NS.etcd_orm.client.write(obj.__name__, '', dir=True, ttl=ttl)
```
r0h4n: if ttl:
   self.client.write(obj.__name__, '', dir=True, ttl=ttl)

GowthamShanmugam: @r0h4n i will change it

Issue #269, opened 2017-03-20
Integration id should not be mandatory parameter for Job


anmolbabu:
Integration id should not be a mandatory parameter for job because:

1. Integration id which is basically cluster id is currently required by commons job framework to be always present.
2. There are integrations like performance-monitoring that can exist even when there's no cluster.

So, integration id should be made optional parameter...

anmolbabu:
@r0h4n I have raised the issue for this...

nthomas-redhat:
Closing the issue as it is already merged

PR #278, opened 2017-03-22
Integration id should not be mandatory parameter for Job

GowthamShanmugam:
tendrl-bug-id: Tendrl/commons#269

### DIFF
--- a/tendrl/commons/jobs/__init__.py
+++ b/tendrl/commons/jobs/__init__.py
@@ -68,10 +68,10 @@ class JobConsumerThread(gevent.greenlet.Greenlet):
                             status=raw_job['status'],
                             payload=json.dumps(raw_job['payload']),
                             errors=raw_job['errors']).save()
-
-                        raw_job['payload']['parameters']['integration_id'] = \
-                            raw_job[
-                            'payload']['integration_id']
+                        if "integration_id" in raw_job['payload']:
+                            raw_job['payload']['parameters']['integration_id'] = \
+                                raw_job[
+                                'payload']['integration_id']
                         raw_job['payload']['parameters']['node_ids'] = raw_job[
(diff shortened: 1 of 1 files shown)


GowthamShanmugam:
@r0h4n @anmolbabu @shtripat @nthomas-redhat please reviewthis

tendrl/commons/jobs/__init__.py
```diff
@@ -68,10 +68,10 @@ def _run(self):
                             status=raw_job['status'],
                             payload=json.dumps(raw_job['payload']),
                             errors=raw_job['errors']).save()
-
-                        raw_job['payload']['parameters']['integration_id'] = \
-                            raw_job[
-                            'payload']['integration_id']
+                        if "integration_id" in raw_job['payload']:
```
shtripat: `raw_job['payload'].keys()` ??

GowthamShanmugam: i will change it

r0h4n: This is ok, no change required

tendrl/commons/jobs/__init__.py
```diff
@@ -68,10 +68,10 @@ def _run(self):
...
+                            raw_job['payload']['parameters']['integration_id'] = \
+                                raw_job[
+                                'payload']['integration_id']
```
r0h4n: Is there any impact on message or job updates if payload['integration_id'] is not present?

GowthamShanmugam: for job updates and message cluster_id is an optional, so i don't think it will affetct

r0h4n: Ok, If you are sure about, This PR looks good

Issue #293, opened 2017-03-23
Add fields `cluster_name` and `cluster_id` under class `tendrl_context`


shtripat:

nthomas-redhat:
Closing the issue as it is already fixed

PR #295, opened 2017-03-23
Added fields cluster_name and cluster_id to TendrlContext class

shtripat:
tendrl-bug-id: Tendrl/commons#293


shtripat:
@anivargi plz take a note of these changes and do the needful in API layer (if any impact at all)

PR #296, opened 2017-03-23
Setting ttl for messages using refresh method

GowthamShanmugam:


GowthamShanmugam:
@shtripat @r0h4n

r0h4n:
Is this verified?

GowthamShanmugam:
@r0h4n i have verified this, refresh will set ttl for already present directory correctly.

Issue #289, opened 2017-03-22
cat should not be added to trusted command list in command utils


anmolbabu:
Instead of doing a cat, the file which is cat can instead be opened through code instead of going through the command utils...

This can potentially impact other modules and hence needs to be carefully undertaken..

GowthamShanmugam:
Patch for this merged successfully.

nthomas-redhat:
Closing the issue as it is already fixed

PR #297, opened 2017-03-23
cat commnads are replace by reading from file

GowthamShanmugam:
tendrl-bug-id: Tendrl/commons#289

### DIFF
--- a/tendrl/commons/objects/memory/__init__.py
+++ b/tendrl/commons/objects/memory/__init__.py
@@ -1,4 +1,6 @@
+from tendrl.commons.event import Event
 from tendrl.commons import etcdobj
+from tendrl.commons.message import ExceptionMessage
 from tendrl.commons.utils import cmd_utils
 
 from tendrl.commons import objects
@@ -24,10 +26,20 @@ class Memory(objects.BaseObject):
                    "Type":      "type"}, ...], ...}
 
         '''
-
-        cmd = cmd_utils.Command("cat /proc/meminfo")
-        out, err, rc = cmd.run()
-        out = str(out)
+        try:
+            out = None
+            with open('/proc/meminfo') as f:
+                out = f.read()
+        except IOError as ex:
+            Event(
+                ExceptionMessage(
+                    priority="error",
+                    publisher=NS.publisher_id,
+                    payload={"message": "Unable to find memory details",
+                             "exception": ex
+                             }
+                )
+            )
 
         if out:
             info_list = out.split('\n')
--- a/tendrl/commons/objects/node_context/__init__.py
+++ b/tendrl/commons/objects/node_context/__init__.py
@@ -1,6 +1,7 @@
 import logging
 import os
 import socket
+import sys
 import uuid
 
 import errno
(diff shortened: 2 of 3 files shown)


tendrl/commons/objects/node_context/__init__.py
```diff
@@ -29,9 +30,17 @@ def __init__(self, machine_id=None, node_id=None, fqdn=None,
...
-        return str(out)
+        try:
+            out = None 
+            with open('/etc/machine-id', 'r') as f:
+                out = f.read().strip('\n') 
+        except IOError as ex:
+            exc_type, exc_value, exc_tb = sys.exc_info()
+            traceback.print_exception(
+                exc_type, exc_value, exc_tb, file=sys.stderr)
+            sys.stderr.write(
```
GowthamShanmugam: Message class required node_context to log the message, so we can't log this exception using message class

GowthamShanmugam:
@shtripat @r0h4n @nthomas-redhat  please review

tendrl/commons/objects/node_context/__init__.py
nthomas-redhat: what is the point here? Ideally you should log using the Event(https://github.com/Tendrl/commons/pull/297/commits/80bc98313d23b8c9afc45bf0404ea9ef0c7d52b0#diff-e3ba3b401bfdec0d0df1d4ee946980abR34)?

GowthamShanmugam: @nthomas-redhat problem is message class needs node_context to log a message using event. But here still node_context is not available in NS, so we can't log the message using message class.

GowthamShanmugam:
@nthomas-redhat no i am not blacklist cat, i will do it

GowthamShanmugam:
@nthomas-redhat i have black listed cat command

tendrl/commons/objects/node_context/__init__.py
GowthamShanmugam: @nthomas-redhat if you see in message class,for each message it assigning node_id automatically,
self.node_id = NS.node_context.node_id
so at this time node id not created so message class wil give error, so we cant use message class for this exception

tendrl/commons/utils/cmd_utils.py
```diff
@@ -9,7 +9,6 @@
 
 SAFE_COMMAND_LIST = [
     "lsblk",
-    "cat",
```
nthomas-redhat: Check all our source and make sure we are not using 'cat' anywhere?
Please confirm

GowthamShanmugam: ok

tendrl/commons/objects/memory/__init__.py
```diff
@@ -24,10 +26,20 @@ def _getNodeMemory(self):
                    "Type":      "type"}, ...], ...}
 
         '''
-
-        cmd = cmd_utils.Command("cat /proc/meminfo")
-        out, err, rc = cmd.run()
-        out = str(out)
+        try:
+            out = None
+            with open('/proc/meminfo', 'r') as f:
```
r0h4n: You dont need 'r' here, it is the default mode

GowthamShanmugam: ok i will change it

GowthamShanmugam:
@r0h4n changes done

PR #321, opened 2017-03-28
Cleanup cat command from code

GowthamShanmugam:
I have already removed cat command from cmd_utils (https://github.com/Tendrl/commons/pull/297/files#diff-216815e55af1fc237caa680457e88c81L12) but again i see cat command in same file so i am removing this, 


GowthamShanmugam:
@r0h4n

r0h4n:
@GowthamShanmugam When are you gonna replace "cat" in other tendrl repos?

PR #337, opened 2017-03-29
Fixing traceback issue in ExceptionMessage class

GowthamShanmugam:

### DIFF
--- a/tendrl/commons/message.py
+++ b/tendrl/commons/message.py
@@ -107,42 +107,51 @@ class Message(object):
 
 
 class ExceptionMessage(Message):
+    # Normal trace will give function calls from try block
+    # but this function will identify traceback from start and
+    # traceback from try block, so it will give traceback like
+    # default python traceback
     def __init__(self, priority, publisher, payload):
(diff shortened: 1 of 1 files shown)


GowthamShanmugam:
Here i am using to traceback one is formatted_stack = traceback.extract_stack()[:-2]
it will give traceback from stack, it give tracebck before try
 formatted_stack = traceback.extract_stack()[:-2]
it will give traceback from stack, it give tracebck before try
for example 

def function1():
    function2()

def function2():
     function3()

def  function3()
    try:
        function4()
    except Exception as ex:
        pass

def function4()
    raise ValueError

here traceback.extract_stack() will give trace back from function1 to function3 it can't give function4 
traceback because it is stack so when control come to except block at that time function4 is removed from stack so upto function3 call only present in stack.

traceback.extract_tb(exc_traceback) will give trace back inside try so it gives function call happened from try. so i am appending both result so this will give complete traceback result.

GowthamShanmugam:
@r0h4n @shtripat please review

PR #339, opened 2017-03-29
Removed socket usage for node-agent log messages

anmolsachan:
tendrl-issue-id: https://github.com/Tendrl/node-agent/issues/355

### DIFF
--- a/tendrl/commons/event.py
+++ b/tendrl/commons/event.py
@@ -3,16 +3,20 @@ from gevent.socket import error as socket_error
 from gevent.socket import timeout as socket_timeout
 import sys
 from tendrl.commons.message import Message
+from tendrl.commons.logger import Logger
 import traceback
 
 
 class Event(object):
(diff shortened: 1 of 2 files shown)


tendrl/commons/event.py
```diff
@@ -3,16 +3,20 @@
...
 
 class Event(object):
     def __init__(self, message, socket_path=None):
-        self.sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
-        self.socket_path = socket_path
-        if self.socket_path is None:
-            self.socket_path = NS.config.data['logging_socket_path']
-        self._write(message)
+        if message.publisher == "node_agent":
+            Logger(message)
```
GowthamShanmugam: problem here is message validation is not happening, so you call message = Message.from_json(json_str) how we used in handler.py

message = Message.from_json(json_str) should comes inside try block

Issue #351, opened 2017-03-30
If ceph allows spaces in cluster name


shtripat:
If not add check in API and tendrl core to not allow spaces in cluster name.

anmolsachan:
This is present in ceph documentation : http://docs.ceph.com/docs/master/install/manual-deployment/  

Cluster Name: Ceph clusters have a cluster name, which is a simple string without spaces. The default cluster name is ceph, but you may specify a different cluster name. Overriding the default cluster name is especially useful when you are working with multiple clusters and you need to clearly understand which cluster your are working with.

shtripat:
With the above definitions, we need to add checks in API layer and in create cluster flow to not allow spaces in cluster name.

PR #354, opened 2017-03-30
Added a check to not allow space char in cluster name

shtripat:
tendrl-bug-id: Tendrl/commons#351


tendrl/commons/flows/create_cluster/__init__.py
```diff
@@ -9,13 +9,32 @@
...
 from tendrl.commons.objects.job import Job
 
 
 class CreateCluster(flows.BaseFlow):
     def run(self):
+        # Check if clusre name contains space char and fail if so
+        if ' ' in self.parameters['TendrlContext.cluster_name']:
+            Event(
+                Message(
+                    priority="info",
```
nthomas-redhat: info?I think this should be an 'error'

shtripat: This info goes as job update because `job_id` and `flow_id` both are present and so would be listed under job updates. Actual error would happen due to `FlowExecutionFailedError`

Issue #264, opened 2017-03-20
Add gluster cluster provisioning logic


shtripat:
Enable gluster cluster creation using gdeploy.

r0h4n:
Is this done?

shtripat:
Create cluster flow is merged already. It works with selinux disabled on the storage nodes at the moment because of authorize_key issue https://github.com/Tendrl/commons/issues/274.
CRUD flows, @nnDarshan is working already.

PR #368, opened 2017-04-03
Provision gluster provisioner after import

nnDarshan:
This patch aims at provisioning the gluster
prvisioner (gdeploy and python-gdeploy) after
the gluster cluster import has happened. This
sets the tag of that node as "gluster/provisioner"
and all the gluster management calls must be
targeted at this node.

tendrl-bug-id: Tendrl/commons#264

### DIFF
--- a/tendrl/commons/flows/import_cluster/__init__.py
+++ b/tendrl/commons/flows/import_cluster/__init__.py
@@ -85,6 +85,25 @@ class ImportCluster(flows.BaseFlow):
             )
         )
 
+        # in case of gluster after import we have to choose one node as
+        # provisioner and insall gdeploy and python-gdeploy in that node.
+        # after that set passwordless ssh between provisioner and other
+        # nodes of cluster.
+        # For this we have a job called setup_gdeploy. The node that picks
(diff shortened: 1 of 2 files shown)


nnDarshan:
@r0h4n @shtripat @nthomas-redhat Please review this.

nnDarshan:
@r0h4n @shtripat @nthomas-redhat When I try to test this, I see that import job is being picked up by multiple node_agents. If that happens, gdeploy-provisioning will not happen in the desired way.

So, we have to make sure that an import job is initially picked up by only one node. Any thoughts on this ?

Or one more option is to separate the gdeploy-provisioning flow from import flow and ask API layer/ UI layer to invoke the gdeploy-provisioning job explicitly after import is successful.

Please provide your inputs

shtripat:
@nnDarshan I am ok with second approach as long as its seamless for end user.

shtripat:
@nnDarshan For approach one to work in clean way we do need global locking mechanism of jobs.

r0h4n:
Isnt this doing the same https://github.com/Tendrl/commons/pull/376  ?

Do we need this PR?

Issue #274, opened 2017-03-21
Authorize_key util is not working when selinux is enabled


GowthamShanmugam:
If selinux is disabled then authorize utils is copying key to particular place without any issue, and it is automatically creating .ssh/authorized_keys  file and folder also.
Actual problem is if selinux is enabled it gives error like ValueError('No JSON object could be decoded',).
first case is:
    if .ssh/authorized_keys file is exist then it is copying key successfully but it give error also. 
second case is:
   if .ssh/authorized_keys file not exist when i run first time it creating authorized_keys but file is empty, second time it copying key successfully also raising error. third time no error came.

GowthamShanmugam:
@shtripat @r0h4n

GowthamShanmugam:
@shtripat @r0h4n

r0h4n:
You are not setting the "manage_dir" flag for the ansible module "authorize_key", you will have to set it along with user and key

https://github.com/ansible/ansible-modules-core/blob/devel/system/authorized_key.py#L50

That should help

GowthamShanmugam:
@r0h4n default manage_dir value is true so no need to give this, i ahve tested with this also

r0h4n:
@GowthamShanmugam Please find root cause of failure of ansible authorize_key module, we cannot hack around our way if the ansible module is faulty. We should file bugs on ansible module if it doesnt behave as it should

GowthamShanmugam:
@r0h4n ok

r0h4n:
Is this issue resolved?

GowthamShanmugam:
_@r0h4n @shtripat it is fixed, we can close this issue

PR #369, opened 2017-04-03
Fix for authorize key module

GowthamShanmugam:
tendrl-bug-id: Tendrl/commons#274

### DIFF
--- a/tendrl/commons/utils/ansible_module_runner.py
+++ b/tendrl/commons/utils/ansible_module_runner.py
@@ -50,6 +50,8 @@ class AnsibleRunner(object):
             raise ValueError
         else:
             self.argument_dict = kwargs
+            self.argument_dict['_ansible_selinux_special_fs'] = \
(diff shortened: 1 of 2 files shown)


GowthamShanmugam:
@r0h4n @shtripat @nnDarshan

shtripat:
@GowthamShanmugam even after this PR merged, I see provisioner asking for confirmation while copying ssh key to other nodes.

GowthamShanmugam:
@shtripat but i am setting environment variable for ansible in https://github.com/Tendrl/node-agent/blob/develop/tendrl-node-agent.service#L8

GowthamShanmugam:
@shtripat then environment variable is in tendrl-node-agent.service, so if we run as a binary then it does not work, if we run as service then it will work fine.

shtripat:
@GowthamShanmugam thanks. that was the issue. it works fine.

PR #376, opened 2017-04-05
Provision gluster provisioner

nnDarshan:
This patch aims at provisioning the gluster
prvisioner (gdeploy and python-gdeploy) before
the gluster cluster import and create flow. This
sets the tag of that node as "gluster/provisioner"
and all the gluster management calls must be
targeted at this node.

tendrl-bug-id: Tendrl/commons#264

### DIFF
--- a/tendrl/commons/flows/utils.py
+++ b/tendrl/commons/flows/utils.py
@@ -4,6 +4,8 @@ import uuid
 from tendrl.commons.event import Event
 from tendrl.commons.message import Message
 from tendrl.commons.objects.job import Job
+from tendrl.commons.utils.ssh import authorize_key
+from tendrl.commons.utils import ansible_module_runner
 
 def to_camel_case(snake_str):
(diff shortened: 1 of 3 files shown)


nnDarshan:
@r0h4n @shtripat @nthomas-redhat PLease review. This is a WIP patch and not tested yet

tendrl/commons/flows/utils.py
```diff
@@ -47,6 +49,49 @@ def ceph_create_ssh_setup_jobs(parameters):
 
 def gluster_create_ssh_setup_jobs(parameters):
     node_list = parameters['Node[]']
+
+    # Install python-gdeploy on the node
```
r0h4n: This function deals with setting up ssh jobs, please move this gdeploy setup in a new function

shtripat: Accept. @nnDarshan plz move to different function.

PR #380, opened 2017-04-11
Fixed loading of python-gdeploy modules

shtripat:

### DIFF
--- a/tendrl/commons/flows/create_cluster/gluster_help.py
+++ b/tendrl/commons/flows/create_cluster/gluster_help.py
@@ -1,4 +1,3 @@
-
 from tendrl.commons.event import Event
 from tendrl.commons.message import Message
 from tendrl.commons.flows.exceptions import FlowExecutionFailedError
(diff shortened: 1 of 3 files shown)


shtripat:

tendrl/commons/flows/create_cluster/utils.py
```diff
@@ -88,6 +88,12 @@ def install_python_gdeploy():
             "Failed to install python-gdeploy"
         )
 
+    # Write the python-gdeploy configuration file
+    config_data = "[python-gdeploy]\n" + \
+        "glusterfs_repo = https://buildlogs.centos.org/centos/7/storage/x86_64/gluster-3.9/"
```
r0h4n: how does this work if i want to use a different glusterfs repo and not use the centos repo

tendrl/commons/flows/create_cluster/gluster_help.py
```diff
@@ -25,6 +24,13 @@ def create_gluster(parameters):
                      }
         )
     )
+
+    # Reimport the python-gdeploy module as it got installed just now
```
r0h4n: instead of doing this, reloading the gluster provisioner plugin would be better

tendrl/commons/flows/create_cluster/utils.py
shtripat: This is WIP PR and ideally this would configurable (may be in node-agent.conf.yaml).
@nnDarshan in case of rhel gluster repos are configured, would it still need to set this configuration in file?

tendrl/commons/flows/create_cluster/gluster_help.py
shtripat: Yes will try that. Just for testing I was doing this.

shtripat: I feel from commons we should not explicitly load a provisioner from node-agent. We should retain importing of individual module of python-gdeploy needed by plugin here.

shtripat:
Verified the changes with create gluster cluster flows. Works as expected and marks one of the cluster nodes a provisioner first. This provisioner is used for creation of the cluster. The created cluster gets imported properly post creation and `tendrl-gluster-integration` run properly on all the gluster nodes.

PR #383, opened 2017-04-12
Correcting exception message in job

GowthamShanmugam:
ExceptionMessage and job updates are different, because exception message have traceback, traceback should not shown to user. Exception message class only tack payload, priority and publisher, otherwise it give keyerror.

Signed-off-by: root <root@dhcp43-207.lab.eng.blr.redhat.com>

### DIFF
--- a/tendrl/commons/jobs/__init__.py
+++ b/tendrl/commons/jobs/__init__.py
@@ -126,14 +126,21 @@ class JobConsumerThread(gevent.greenlet.Greenlet):
                             Event(
                                 ExceptionMessage(
                                     priority="error",
-                                    job_id=job.job_id,
-                                    flow_id = the_flow.parameters['flow_id'],
                                     publisher=NS.publisher_id,
                                     payload={"message": "error",
                                              "exception": e
                                              }
                                 )
                             )
+                            Event(
+                                Message(
+                                    job_id=job.job_id,
+                                    flow_id = the_flow.parameters['flow_id'],
+                                    priority="error",
+                                    publisher=NS.publisher_id,
+                                    payload={"message": "Job failed"}
+                                )
+                            ) 
                             try:
                                 NS.etcd_orm.client.write(job_status_key, "failed", prevValue="processing")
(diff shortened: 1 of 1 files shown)


GowthamShanmugam:
@shtripat @r0h4n

tendrl/commons/jobs/__init__.py
```diff
@@ -126,14 +126,21 @@ def _run(self):
...
                                              }
                                 )
                             )
+                            Event(
+                                Message(
+                                    job_id=job.job_id,
+                                    flow_id = the_flow.parameters['flow_id'],
+                                    priority="error",
+                                    publisher=NS.publisher_id,
+                                    payload={"message": "Job failed"}
```
r0h4n: How is "Job failed" explaining anything to the user? Add the e.msg from exception caught

tendrl/commons/jobs/__init__.py
```diff
@@ -126,14 +126,21 @@ def _run(self):
                             Event(
                                 ExceptionMessage(
                                     priority="error",
-                                    job_id=job.job_id,
```
r0h4n: ExceptionMessage is a subclass of Message, it can take job_id and flow_id as input

tendrl/commons/jobs/__init__.py
GowthamShanmugam: i will change it

tendrl/commons/jobs/__init__.py
GowthamShanmugam: but it will display trackback also is that ok?

GowthamShanmugam: if we use it is stored as a job_update then it never present in event table, because event and job updates are different.

GowthamShanmugam: if job_id and flow_id is present then it will treat as job_update only not as event

r0h4n: ok

PR #404, opened 2017-04-20
Added output field for job

shtripat:
If some job wants to update the output of the job inline
this field could be used.

### DIFF
--- a/tendrl/commons/objects/definition/master.yaml
+++ b/tendrl/commons/objects/definition/master.yaml
@@ -511,6 +511,9 @@ namespace.tendrl:
         children:
           help: "ID of child jobs created by this job"
           type: List
+        output:
+          help: output of the job execution
+          type: dict
       enabled: true
       list: /queue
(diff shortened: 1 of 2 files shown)


shtripat:
This PR is sent to enable output population inline within job if some job wants to.

@brainfunked this would help saving calculated journal_mapping for ceph use case as suggested in https://github.com/Tendrl/node-agent/pull/371#pullrequestreview-33480058

r0h4n:
Would this mean that all atoms inside a specific job can overwrite/append the Job.output field ?

shtripat:
With this yes. Do you feel we should have output as dict with atom name as keys so that multiple atoms can write their output to the same field?

r0h4n:
The API understands atoms run for a Job by its definitions. I would prefer a "JobOutput" object (takes job id as init param) which can be has attributes as atom names and value as the atom output, the flow can also add it output via its name in JobOutput object

PR #410, opened 2017-04-20
Set empty dict as `output` for jobs

shtripat:
This would be helpful for flows/atoms for updating the output
data.

### DIFF
--- a/tendrl/commons/jobs/__init__.py
+++ b/tendrl/commons/jobs/__init__.py
@@ -55,6 +55,11 @@ class JobConsumerThread(gevent.greenlet.Greenlet):
                             pass
 
                         job = Job(job_id=jid).load()
+
+                        # Set the empty dict as output for the job
+                        job.output = json.dumps({})
+                        job.save()
+
                         raw_job = {}
                         raw_job["payload"] = json.loads(job.payload.decode('utf-8'))
                     except etcd.EtcdKeyNotFound:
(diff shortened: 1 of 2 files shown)


tendrl/commons/objects/job/__init__.py
```diff
@@ -22,6 +22,7 @@ def __init__(self, job_id=None, payload=None,
     def save(self):
         super(Job, self).save()
         payload = json.loads(self.payload)
+        output = json.loads(self.output)
         if "parent" in payload:
```
r0h4n: Not required

tendrl/commons/jobs/__init__.py
```diff
@@ -55,6 +55,11 @@ def _run(self):
                             pass
 
                         job = Job(job_id=jid).load()
+
+                        # Set the empty dict as output for the job
+                        job.output = json.dumps({})
+                        job.save()
```
r0h4n: Just dict() , remove json.dumps

tendrl/commons/objects/job/__init__.py
shtripat: Done

tendrl/commons/jobs/__init__.py
shtripat: Just `dict()` doesnt work. It works with `dict(_None=None)`

PR #417, opened 2017-04-26
Corrected reading of major minor version for sds

shtripat:
Sometime the version no could be of form `10.2.7-8.el7cp`.
The current way of reading the major, minor versions and
release no is not correct.

Anyway we need only major and minor version details so changed
to read only the required details.


shtripat:

PR #420, opened 2017-04-26
Flag to convert ansible command module to shell module

GowthamShanmugam:
Using  _uses_shell flag command module can support the shell commands also

Signed-off-by: root <root@dhcp43-149.lab.eng.blr.redhat.com>

### DIFF
--- a/tendrl/commons/utils/cmd_utils.py
+++ b/tendrl/commons/utils/cmd_utils.py
@@ -29,10 +29,10 @@ class UnsupportedCommandException(Exception):
 
 
 class Command(object):
-    def __init__(self, command):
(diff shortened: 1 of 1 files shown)


GowthamShanmugam:
@shtripat @r0h4n @nnDarshan  @nthomas-redhat I have used shell ansible module to run shell command, but as per ansible runner implementation in common shell module is not supported. i have go through the ansible code there i found  _uses_shell, when we give shell use shell module via ansible playbook, it actually call command module with extra _uses_shell parameter. so i have added his flag in cmd_utils as optional.

GowthamShanmugam:
@shtripat yes i verified

Issue #180, opened 2017-03-02
Unit tests fail for py27 enviroment because of importing non existing modules


mbukatov:
Two tests from py27 test run fails:

```
tendrl/commons/tests/test_base_flow.py::TestBaseFlow::test_run ERROR
tendrl/commons/tests/test_base_flow.py::TestBaseFlow::test_extract_atom_details ERROR
```

Errors collected by pytest show importing modules which doesn't exist, eg:

```
ImportError while importing test module '/home/travis/build/Tendrl/commons/tendrl/commons/tests/test_commons.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
tendrl/commons/tests/test_commons.py:12: in <module>
    from tendrl.commons.etcdobj import etcdobj
E   ImportError: cannot import name etcdobj
```

See [full log from Travis CI](https://travis-ci.org/Tendrl/commons/jobs/206888065).

rishubhjain:
@r0h4n @mbukatov  I guess this issue is resolved and must be closed, If there is something left then I would like to contribute.

mbukatov:
@rishubhjain @r0h4n I rechecked the state of unit tests and noticed that there are no unit tests in `tendrl-commons`. So yes, this issue is no longer valid.

rishubhjain:
Great, we can close the issue :)

r0h4n:
@rishubhjain Your PR provides unit tests, I have linked this issue to the PR

@mbukatov please review https://github.com/Tendrl/commons/pull/433

rishubhjain:
There is already an issue raised for unit tests for which I guess https://github.com/Tendrl/commons/pull/433 will be valid for (https://github.com/Tendrl/commons/issues/85) .
@r0h4n Is there a need to keep this issue alive for the same PR?. I feel this issue is bit out of topic w.r.t to PR

mbukatov:
@rishubhjain your comment about your pull request being off topic seems valid to me, but it's for @r0h4n to settle this in the end - if he thinks that shifting the scope of this issue be to more general based just on the title is better with respect to the issue management, I don't mind

r0h4n:
@rishubhjain lets use this to track the unit tests PR, we dont want to spend a lot of time duplicating bug reports

rishubhjain:
@r0h4n cool, will continue with this

rishubhjain:
[...]

PR #433, opened 2017-05-02
Unit Test and Code optimization

rishubhjain:
Tried to remove redundancy.
A separate function(def _log()) could be created for LOG messages to remove repetition of code(try except block). [ LEFT FOR DISCUSSION]


tendrl/commons/__init__.py
```diff
@@ -1,11 +1,9 @@
 import __builtin__
-import importlib
+from importlib import import_module
 import inspect
-import pkgutil
-
-import maps
+from pkgutil import walk_packages
```
r0h4n: The code is more readable when we call say "pkgutil.walk_packages" than just "walk_packages". So keep the old import style.

Also, "walk_packages" is called in this file, it seems you missed out removing pkgutil from the actual function call

tendrl/commons/__init__.py
```diff
@@ -579,10 +577,10 @@ def get_obj_flow_definition(self, obj_name, flow_name):
                               type=raw_flow['type'],
                               uuid=raw_flow['uuid'])
 
+        raw_atom = obj_def.atoms[atom_name]
```
r0h4n: How is this going to be called after you return at L571

r0h4n:
Also, you can go ahead and define the a module log_utils.py and implement the log (not _log) function in that

https://github.com/Tendrl/commons/tree/develop/tendrl/commons/utils

tendrl/commons/__init__.py
rishubhjain: ok sure

tests/test_init.py
```diff
@@ -0,0 +1,96 @@
+import unittest
```
r0h4n: All unit tests should be located under  eg: "tendrl/commons/tests/test_init.py"

for another example say (tendrl/commons/manager/__init__.py) the tests should be at tendrl/commons/tests/manager/test_init.py

tendrl/commons/tests/test_init.py
```diff
@@ -0,0 +1,159 @@
...
+    #TO-DISCUSS : object_list is hard coded and might change in future
+    assert ret == objects_list
+
+
+#Testing _get_object
+def test_get_object():
+    tendrlNS = TendrlNS()
+    ret = tendrlNS._get_object("Service")
+    assert (inspect.isclass(ret)) == True
+    assert (issubclass(ret, objects.BaseObject)) == True
```
r0h4n: Would be good if the object attributes are compared with their actual definitions and the test should fail if they dont match, 

do this for objects that are non-internal (i.e. objs that dont have "internal" flag)

tendrl/commons/utils/log_utils.py
```diff
@@ -0,0 +1,26 @@
+"""
```
r0h4n: remove this file, it is now in a different PR

tendrl/commons/tests/test_init.py
```diff
@@ -0,0 +1,159 @@
...
+def test_list_modules_in_package_path():
+
+    tendrlNS = TendrlNS("tendrl","tendrl.commons")
+    modules = [('cluster_node_context', 'tendrl.commons.objects.cluster_node_context'), ('cluster_tendrl_context', 'tendrl.commons.objects.cluster_tendrl_context'),\
+              ('cpu', 'tendrl.commons.objects.cpu'),('definition', 'tendrl.commons.objects.definition'), ('detected_cluster', 'tendrl.commons.objects.detected_cluster'),\
+              ('disk', 'tendrl.commons.objects.disk'), ('file','tendrl.commons.objects.file'), ('job','tendrl.commons.objects.job'),\
+              ('memory', 'tendrl.commons.objects.memory'), ('node', 'tendrl.commons.objects.node'), ('node_context', 'tendrl.commons.objects.node_context'),\
+              ('node_network', 'tendrl.commons.objects.node_network'), ('os', 'tendrl.commons.objects.os'), ('platform', 'tendrl.commons.objects.platform'),\
+              ('service', 'tendrl.commons.objects.service'), ('tendrl_context', 'tendrl.commons.objects.tendrl_context')]
+    ns_objects_path = "/home/risjain/Tendrl/commons-develop/tendrl/commons/objects"
```
rishubhjain: Done

tendrl/commons/tests/test_init.py
```diff
@@ -0,0 +1,159 @@
+import py.test
```
rishubhjain: Done

tendrl/commons/tests/test_init.py
```diff
@@ -0,0 +1,159 @@
...
+    ret = tendrlNS._list_modules_in_package_path("test","test")
+    assert len(ret) == 0
+
+
+#Testing _register_subclasses_to_ns
+def test_register_subclasses_to_ns():
+
+    tendrlNS = TendrlNS()
+    tendrlNS._register_subclasses_to_ns()
+    #assert getattr(NS,)
```
rishubhjain: Done

tendrl/commons/utils/log_utils.py
rishubhjain: Done

tendrl/commons/tests/test_init.py
```diff
@@ -0,0 +1,159 @@
...
+        tendrlNS._add_object(obj_name,obj_cls[1])
+        break
+
+    assert isinstance(getattr(current_ns.objects,"_test_obj")['atoms'],maps.NamedDict)
+    assert isinstance(getattr(current_ns.objects,"_test_obj")['flows'],maps.NamedDict)
+
+
+#Testing _get_objects
+def test_get_objects():
+    objects_list = ['Service', 'Platform', 'Memory', 'ClusterTendrlContext', 'Node', 'NodeContext',\
```
rishubhjain: I have made the changes according to what I understood which is creating a directory hierarchy(not including files as it would just be duplication) and using that hierarchy to generate dummy testing data for the functions like get_objects(), I am not sure but were you instructing me to create a fixture where I would create a complete NS (populating it with objects and flows) and then using that NS?

tendrl/commons/tests/test_init.py
rishubhjain: I have compared returned objects attribute with definition from master.yaml file. For now I have just checked for NodeNetwork. Do check if this is the right way to do it

r0h4n:
`tendrl/commons/__init__.py                                                329    113     86     20    65%
`

Coverage still remains at 65%, are you blocked on anywhere @rishubhjain ?

tendrl/commons/fixtures/nodecontext.py
```diff
@@ -0,0 +1,4 @@
+class NodeContext():
+
+    def __init__(self,*args):
+        NodeContext = "Test Code"
```
r0h4n: use actual node_id, machine_id and all other attributes (assign them fake values) from the original NodeContext , TendrlContext classes

r0h4n:
Fix the review comments and issues mentioned at https://www.codacy.com/app/tendrl-bot/commons/pullRequest?prid=631578

Rest looks good to me, nice work!

r0h4n:
tendrl-bug-id: Tendrl/commons#180

@mbukatov please review this PR, these are the unit tests

mbukatov:
Great to see this effort. I'm going to add few comments.

rishubhjain:
@mbukatov thanks, I am sure I might have missed few things, would make the required changes

mbukatov:
@rishubhjain I see one failing case: `tendrl/commons/tests/test_init.py::test_get_objects`

```
    def test_get_objects():
        path = os.path.join(os.path.dirname(
                            os.path.dirname(os.path.abspath(__file__))), "objects")
        objects_list = [d.title().replace('_', '') for d in os.listdir(path)
                        if os.path.isdir(os.path.join(path, d))]
        tendrlNS = init()
        ret = tendrlNS._get_objects()
        assert isinstance(objects_list, list)
        assert ret is not None
        # TO-DISCUSS : object_list is hard coded and might change in future
>       assert set(ret) == set(objects_list)
E       assert set(['BlockDe...luster', ...]) == set(['Atoms', ...nition', ...])
E         Extra items in the right set:
E         'Atoms'
E         Full diff:
E         - set(['BlockDevice',
E         + set(['Atoms',
E         +      'BlockDevice',
E         'ClusterNodeContext',
E         'ClusterTendrlContext',
E         'Cpu',
E         'Definition',
E         'DetectedCluster',
E         'Disk',
E         'Job',
E         'Memory',
E         'Node',
E         'NodeContext',
E         'NodeNetwork',
E         'Os',
E         'Platform',
E         'Service',
E         'TendrlContext'])

tendrl/commons/tests/test_init.py:159: AssertionError
```

Do you still work on this one? Based on a TODO node there, maybe a discussion is still needed?

mbukatov:
This is comment about sharing [pytest fixtures](https://docs.pytest.org/en/latest/fixture.html) and expected test structure.

I recently noticed [how Sphinx project manages fixture code](https://github.com/sphinx-doc/sphinx/blob/master/sphinx/testing/__init__.py) - which is important because setup of a sphinx project from unit test is quite complicated and one can't just import pytest conftest.py module, because it's outside of application module - and thought that it could be a good approach for Tendrl as well, since fixtures defined in tendrl-commons would be useful in unit tests of other Tendrl components.

But then I realized that Tendl uses different approach: unit tests are inside application module, so we may need to check [Tests as part of application code](https://docs.pytest.org/en/latest/goodpractices.html#tests-as-part-of-application-code).

I will think it over and suggest what a best approach would be here. I would like to follow expected practices from pytest point of view and make the test fixtures reusable for other tendrl projects using tendrl-commons.

rishubhjain:
@mbukatov thanks for the review
Regarding the assertion error: I will check again for the reason for the fail in the updates code,
Regarding the To-Discuss note , I discussed it with Rohan and it should be hard coded.
If you have any other suggestions then we could discuss.

Regarding the change in fixtures/restructuring test code, may be we can discuss on the best approach with @r0h4n and come up with a final approach

r0h4n:
waiting for @mbukatov to comment in detail, tests look good for me, suggestions can be taken in another PR

rishubhjain:
@r0h4n Done, https://travis-ci.org/Tendrl/commons/jobs/237310879#L290

rishubhjain:
@mbukatov I checked for the error that you just mentioned in the above comments, I am not able to reproduce the same error, And I am getting no error in the latest code (https://travis-ci.org/Tendrl/commons/jobs/237310879#L290) .

Do you have the updated code? Also do check if there is unwanted Atoms folder in tendrl/commons/objects in your local code.

tendrl/commons/fixtures/config.py
```diff
@@ -0,0 +1,5 @@
+class Config():
```
r0h4n: please change the location of fixtures to tendrl/commons/tests/fixtures

rishubhjain:
@r0h4n Done https://github.com/Tendrl/commons/pull/433/commits/53263491ee1504b885cfbcd04977587ca2fec552

PR #434, opened 2017-05-03
Added condition to check messages comes with empty cluster_id

GowthamShanmugam:
some places empty cluster id is used before it assign for example : https://github.com/Tendrl/node-agent/blob/develop/tendrl/node_agent/provisioner/gluster/plugins/gdeploy.py, This time cluster_id is not available. So it store messages like /clusters/messages/.

Signed-off-by: root <root@dhcp43-207.lab.eng.blr.redhat.com>


GowthamShanmugam:
@r0h4n @nthomas-redhat @anivargi please review

PR #437, opened 2017-05-05
Utility function to log errors/info etc

rishubhjain:


tendrl/commons/utils/log_utils.py
```diff
@@ -0,0 +1,26 @@
+"""
+Handles logging functionality.
```
anmolsachan: Logging is done at many places and calling another function will increase the overhead I guess. Second at each logging instance the payload structure is not same. Third this could have been done inside the event class but it was not as it was not recommended in the spec discussion.

tendrl/commons/utils/log_utils.py
```diff
@@ -0,0 +1,26 @@
...
+
+def log(log_priority, publisher_id, msg):
+    """
+    Function used for logging errors/output/info.
+    Args:
+        log_priority : Priority of the Log Message
+        publisher_id : Id of publisher
+        msg          : Message to be logged
+    """
+    try:
```
anmolsachan: Try Except block is not required everywhere. It was used at the places which were invoked before Namespace NS was created.

tendrl/commons/utils/log_utils.py
```diff
@@ -0,0 +1,26 @@
+"""
+Handles logging functionality.
+"""
+import sys
+from tendrl.commons.event import Event
+from tendrl.commons.message import Message
+
+
+def log(log_priority, publisher_id, msg):
```
anmolsachan: Rather than passing 'msg' you should pass the payload as it is not the same everywhere. It is different for jobs, logs etc

tendrl/commons/utils/log_utils.py
```diff
@@ -0,0 +1,26 @@
...
+    try:
+        Event(
+            Message(
+                priority=log_priority,
+                publisher=publisher_id,
+                payload={"message": msg}
+            )
+        )
+    except KeyError:
+            sys.stdout.write(msg)
```
anmolsachan: We write to stdout and stderr both based on the priority. U will have to keep a check which logs to stderr if the priority is error and otherwise to stdout.

tendrl/commons/utils/log_utils.py
r0h4n: Calling a util function for logging is ok. It helps reduce the clutter in actual code (where we create event class etc).

@anmolsachan please ensure this util is generic enough to handle all our logging scenarios. Provide comments related to that here

tendrl/commons/utils/log_utils.py
r0h4n: Try except is ok here, since we do have to handle the scenario you mentioned @anmolsachan

tendrl/commons/utils/log_utils.py
anmolsachan: I feel we can modify the event class only and put this logic in that. We will have to call the event class only then.

anmolsachan: I think we can go ahead with log_util as rohan suggested.

tendrl/commons/utils/log_utils.py
rishubhjain: I have made required changes in the log_utils file, it should be now able to handle dynamic payload format, please do review it and let me know if any changes are required

tendrl/commons/utils/log_utils.py
rishubhjain: Done, Please check !

tendrl/commons/utils/log_utils.py
```diff
@@ -0,0 +1,31 @@
...
+                Message(
+                    priority=log_priority,
+                    publisher=publisher_id,
+                    payload=log_payload
+                )
+            )
+        else:
+            raise Exception('Publisher Id not found')
+    except:
+            message = log_payload.get("message","Meesage Not Found") or log_payload.get("Message","Meesage Not Found")
```
anmolsachan: I think rather than using log_payload.get,  log_payload["message"] should be used as each log should have a massage. If it is not there then that might be an error, as then there will be no point of logging. Also I think only lower case should be checked as everywhere it is lower. If you find capital then you can change it.

tendrl/commons/utils/log_utils.py
```diff
@@ -0,0 +1,31 @@
...
+            if log_priority.lower() == "error":
```
anmolsachan: I think only lower case should be checked as everywhere it is lower. If you find capital then you can change it. You can refer to https://github.com/Tendrl/commons/blob/develop/tendrl/commons/logger.py . The mappings are in lower case only

anmolsachan:
Please rebase your branch on top of develop.

tendrl/commons/utils/log_utils.py
rishubhjain: log_payload.get() has been used in the except clause where the message is supposed to be printed, I have given this particular condition so that in case someone misses out the message then this function should not throw error

tendrl/commons/utils/log_utils.py
rishubhjain: Here I was trying to indicate that there might be a need to handle all types of cases internally, I wanted to put this up for discussion, whether to  handle all negitive cases or not

tendrl/commons/utils/log_utils.py
anmolsachan: Yeah I get that this is not wrong. My thought was if someone misses out the message then it should throw error and it should be fixed there itself. It is not the kind of exception that should be handled.

rishubhjain: ok, I will make the required changes,

tendrl/commons/utils/log_utils.py
```diff
@@ -0,0 +1,31 @@
...
+        publisher_id : Id of publisher (mandatory)
+        log_payload  : Payload can contain parameters like message that is to be logged
+    """
+    try:
+        if publisher_id != None:
+            Event(
+                Message(
+                    priority=log_priority,
+                    publisher=publisher_id,
+                    payload=log_payload
```
anmolsachan: Other than the given fields above(priority, publisher, payload) ,more fields are to be taken care of for job updates. You can find them here https://github.com/Tendrl/commons/blob/develop/tendrl/commons/message.py#L33

tendrl/commons/__init__.py
```diff
@@ -23,37 +23,14 @@ def __init__(self, ns_name="tendrl", ns_src="tendrl.commons"):
...
-                Message(
-                    priority="info",
-                    publisher=NS.publisher_id,
-                    payload={"message": "namespace.%s created!" % self.ns_name
-                             }
-                )
-            )
-        except KeyError:
-            sys.stdout.write("namespace.%s created!\n" % self.ns_name)
+	Logger.log("info",NS.get("publisher_id", None),message="namespace.%s created!" % self.ns_name)
```
anmolsachan: Other than the given fields above(priority, publisher, payload) ,more fields are to be taken care of for job updates. You can find them here https://github.com/Tendrl/commons/blob/develop/tendrl/commons/message.py#L33

tendrl/commons/utils/log_utils.py
```diff
@@ -0,0 +1,31 @@
+""" Handles logging functionality. """
+import sys
+from tendrl.commons.event import Event
+from tendrl.commons.message import Message
+
+
+def log(log_priority,publisher_id,**log_payload):
```
GowthamShanmugam: Here the problem is message framework is identifying caller, line number, function names from stack, so it will give wrong info for that fields, Better identify those details inside this new function and pass to the message framework.  And don't for get to delete the code for find caller details in message framework.  

caller = getframeinfo(stack()[1][0])
 self.caller = {"filename": caller.filename,
                      "line_no": caller.lineno,
                       "function": caller.function}
add these lines in your function and pass these details to message framework, and remove these code from message framwork.

please check the parameters, i this it will not work for job_updates, you have to add job_id, flow_id, cluster_id, parent these details also in function parameter.   

use parameter like 
 def __init__(self, priority, publisher, payload, job_id=None,
                 flow_id=None, parent_id=None, cluster_id=None,
                 message_id=None, timestamp=None, node_id=None,
                 caller=None)

rishubhjain: Done as per what was required, I wanted to know why handling caller = None was important in the log function itself?

tendrl/commons/__init__.py
rishubhjain: Thanks for the points, I have made the required changes,  The solution is pretty naive, I am working on making it more generic

tendrl/commons/utils/log_utils.py
rishubhjain: Done

tendrl/commons/utils/log_utils.py
rishubhjain: Getting all the parameters one by one is quite naive although I have used that for now, but I am working on more generic way of solving the issue, will update soon

GowthamShanmugam: problem is when we call message call class with some value at that time caller is not required in __init__ param, but when message is received by handler via socket then it will validated using same message class, at that time again __init__ function is called from from_json  function at that time caller have some value, __inti__ function should have caller has parameter. so i gave that as optional.

now you are passing caller details from your function so it will always come so you all change remove that option thing.

tendrl/commons/utils/log_utils.py
```diff
@@ -0,0 +1,41 @@
...
+        caller=None, *args, **kwargs):
+    """
+    Function used for logging errors/output/info.
+    Args:
+        log_priority [Type : String]: Priority of the Log Message (error/info)
+        publisher_id [Type : Integer] : Id of publisher (mandatory)
+        log_payload [Type: Dict] : Payload can contain /
+                                   parameters like message that is to be logged
+    """
+    if caller is None:
```
GowthamShanmugam: this if condition is not required, caller is identified and passed from this function only so no need

tendrl/commons/utils/log_utils.py
```diff
@@ -0,0 +1,41 @@
...
+                      "line_no": caller.lineno,
+                      "function": caller.function}
+    try:
+        if publisher_id is not None:
+            Event(
+                Message(
+                    log_priority, publisher_id, log_payload,
+                    job_id, flow_id, parent_id, cluster_id, message_id,
+                    timestamp, node_id, caller, args, kwargs
+                )
```
GowthamShanmugam: here also remove args, kwargs, message_id, node_id, timestamp
here caller is required so caller is correct

tendrl/commons/utils/log_utils.py
```diff
@@ -0,0 +1,41 @@
...
+        log_payload [Type: Dict] : Payload can contain /
+                                   parameters like message that is to be logged
+    """
+    if caller is None:
+            caller = getframeinfo(stack()[1][0])
+            caller = {"filename": caller.filename,
+                      "line_no": caller.lineno,
+                      "function": caller.function}
+    try:
+        if publisher_id is not None:
```
GowthamShanmugam: you can add this check in https://github.com/rishubhjain/commons/blob/c230cd8ec64390af84120e1a71364085fd7069d8/tendrl/commons/message.py#L75

like
 if publisher_id is None:
    return False

tendrl/commons/utils/log_utils.py
```diff
@@ -0,0 +1,41 @@
...
+                    timestamp, node_id, caller, args, kwargs
+                )
+            )
+        else:
+            raise Exception('Publisher Id not found')
+    except:
+            if log_priority.lower() == "error":
+                sys.stderr.write(log_payload.get("message"))
+            else:
+                sys.stdout.write(log_payload.get("message"))
```
GowthamShanmugam: This exception is not required, you can remove.  Because if you add  condition like
if publisher_id is None:
   return False

In https://github.com/rishubhjain/commons/blob/c230cd8ec64390af84120e1a71364085fd7069d8/tendrl/commons/message.py#L75

then it will stored as a Invalid message in etcd automatically.

tendrl/commons/utils/log_utils.py
rishubhjain: Thanks for the explanantion, so now the caller parameter is not optional in the message.py?

tendrl/commons/utils/log_utils.py
rishubhjain: Done :)

tendrl/commons/utils/log_utils.py
rishubhjain: Done :)

tendrl/commons/utils/log_utils.py
rishubhjain: Done :)

tendrl/commons/utils/log_utils.py
rishubhjain: Done

r0h4n:
@rishubhjain @GowthamShanmugam @anmolsachan Please verify (run the node-agent with this code and ensure etcd contains messages) this PR. Rest looks good

GowthamShanmugam:
@r0h4n yes it will broke, because now caller identified from log_utils function but node agent log messages are directly calling message framework, so there caller is always none. so messages will pushed in etcd but when messages pushed in to syslog at that time it will fail. Anyway i handling https://github.com/Tendrl/node-agent/blob/develop/tendrl/node_agent/message/handler.py#L43 keyerror keyerror check. so it never broke.

But in etcd caller info will come none. He have to change all log messages based on this util.

GowthamShanmugam:
@rishubhjain for now you leave caller as optional, when you change all the log message at that time you just change is to non optional.

tendrl/commons/utils/log_utils.py
```diff
@@ -0,0 +1,41 @@
...
+import sys
+from tendrl.commons.event import Event
+from tendrl.commons.message import Message
+from inspect import getframeinfo
+from inspect import stack
+
+def log(log_priority, publisher_id, log_payload, job_id=None,
+        flow_id=None, parent_id=None, cluster_id=None,
+        message_id=None, timestamp=None, node_id=None,
+        caller=None, *args, **kwargs):
```
rishubhjain: So here I should keep caller =None for now..right?

tendrl/commons/utils/log_utils.py
rishubhjain: And I guess I should check if caller is None, and pass the caller to the Message function as well?

r0h4n:
@GowthamShanmugam @rishubhjain @anmolsachan  I am asking you folks to test whether log messages in tendrl/commons/__init__.py are reflected in etcd and syslog with this new  PR

GowthamShanmugam:
@r0h4n ok Done

tendrl/commons/utils/log_utils.py
GowthamShanmugam: no here caller is not required
remove caller, node_id, timestamp, message_id and args, kwargs

tendrl/commons/utils/log_utils.py
```diff
@@ -0,0 +1,36 @@
...
+        log_priority [Type : String]: Priority of the Log Message (error/info)
+        publisher_id [Type : Integer] : Id of publisher (mandatory)
+        log_payload [Type: Dict] : Payload can contain /
+                                   parameters like message that is to be logged
+    """
+    caller_details = getframeinfo(stack()[1][0])
+    caller_details = {"filename": caller_details.filename,
+                      "line_no": caller_details.lineno,
+                      "function": caller_details.function}
+    try:
```
anmolsachan: Is the try-except block required now ?

tendrl/commons/utils/log_utils.py
```diff
@@ -0,0 +1,36 @@
...
+        Event(
+            Message(
+                log_priority, publisher_id, log_payload,
+                job_id, flow_id, parent_id, cluster_id,
+                caller=caller_details
+            )
+        )
+    except:
+            if log_priority.lower() == "error":
```
anmolsachan: If you are checking for the upper case , then in Message also pass it as log_priority.lower()

r0h4n:
@anmolsachan @GowthamShanmugam @rishubhjain are you guys done with this one?

tendrl/commons/utils/log_utils.py
rishubhjain: @GowthamShanmugam @r0h4n validate function in message.py takes a generic approach to all kind of error(ex: wrong publisher id, wrong type etc), Should we not write appropriate error messages?

r0h4n: We should, please make those changes

rishubhjain:
@r0h4n The code works fine, one possible inclusion would be reporting error messages according to the errors. This could also be done as another PR

anmolsachan:
@r0h4n The problem is before the NS is created the publisher is being passed as None because of which it goes as an invalid message. That has to be handled.

GowthamShanmugam:
@r0h4n @anmolsachan message inside __init__is not present in etcd. we can't log also a invalid message because at log time we are using publisher _id for log message so invalid log message also will give error.

GowthamShanmugam:
@anmolsachan is correct, in valid message cant come in etcd

GowthamShanmugam:
@rishubhjain as we discussed add caller for invalid message, other wise this patch is LGTM

anmolsachan:
@rishubhjain I Think we have to make a change here ( https://github.com/Tendrl/commons/blob/develop/tendrl/commons/message.py#L71 ). As the messages goes invalid due to None publisher_id and here Message object instantiation fails due to to changes u have made to caller . i think here you will have to handle that and call your Logger.log() here or handle caller explicitly

rishubhjain:
@GowthamShanmugam I have made the necessary changes as we have discussed.
@anmolsachan The change in message will require some formatting in the structure of message passed, I have discussed it with gowtham and will do it, for now it may be a secondary task.

rishubhjain:
@r0h4n verified and good to go

GowthamShanmugam:
LGTM

PR #439, opened 2017-05-05
Added Expand gluster cluster feature

nnDarshan:
This patch adds flow for expanding an existing gluster
cluster by setting up the new nodes and adding them to
cluster and importing them


nnDarshan:
@rohan @nthomas-redhat @shtripat WIP patch for expanding gluster cluster. NOT VERIFIED

Please have a look at the approach

r0h4n:
could you also pull in changes from https://github.com/Tendrl/node-agent/pull/402

nnDarshan:
@r0h4n shubhendu is back and he said he would take care of it

nnDarshan:
Have verified half(except syncing/importing of newly added nodes) of it

nnDarshan:
@r0h4n have verified this please have a look

tendrl/commons/flows/expand_cluster/__init__.py
```diff
@@ -0,0 +1,131 @@
+# flake8: noqa
```
nnDarshan: Since this file has references to NS. without this PEP8 will fail

tendrl/commons/flows/expand_cluster/__init__.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+        # Wait till detected cluster in populated for nodes
+        all_nodes_have_detected_cluster = False
+        while not all_nodes_have_detected_cluster:
+            all_status = []
+            for node in self.parameters['Node[]']:
+                try:
+                    NS.etcd_orm.client.read("/nodes/%s/DetectedCluster" % node)
+                    all_status.append(True)
```
nnDarshan: ack

tendrl/commons/flows/expand_cluster/gluster_help.py
```diff
@@ -0,0 +1,69 @@
...
+        )
+
+    Event(
+        Message(
+            job_id=parameters['job_id'],
+            flow_id=parameters['flow_id'],
+            priority="info",
+            publisher=NS.publisher_id,
+            payload={"message": "Expanded Gluster Cluster %s" %
+                                parameters['TendrlContext.integration_id']
```
nnDarshan: ack

tendrl/commons/flows/expand_cluster/utils.py
```diff
@@ -0,0 +1,58 @@
+import copy
```
nnDarshan: This is slightly different, in the sense that during expand we need not write ssh-keys to authorized_key file of current node as current node is already a cluster node and this would have already happened.

If you feel that, the create_cluster utils has to be used, we will have to add a flag in create cluster util to achieve this. Please let me know which one is better.

tendrl/commons/flows/import_cluster/__init__.py
```diff
@@ -23,110 +23,110 @@ def run(self):
...
+                            'nodes/%s/TendrlContext/integration_id' % entry
                         )
-                    )
-
-                    if _integration_id.value != "":
                         Event(
                             Message(
                                 job_id=self.job_id,
                                 flow_id = self.parameters['flow_id'],
                                 priority="info",
```
nnDarshan: ack

nnDarshan:
@r0h4n Please have a look. Have updated the PR

tendrl/commons/flows/expand_cluster/utils.py
r0h4n: Isnt this doing the same as you explained https://github.com/Tendrl/commons/blob/develop/tendrl/commons/flows/create_cluster/utils.py#L140

tendrl/commons/flows/expand_cluster/__init__.py
r0h4n: We have to fix PEP8 if the checks fail

Also NS will be considered as a builtin, are you sure you are running the correct command for pep8? "tox -epep8" ?

https://github.com/Tendrl/commons/blob/develop/tox.ini#L74

tendrl/commons/flows/expand_cluster/utils.py
nnDarshan: https://github.com/Tendrl/commons/blob/develop/tendrl/commons/flows/create_cluster/utils.py#L125 to https://github.com/Tendrl/commons/blob/develop/tendrl/commons/flows/create_cluster/utils.py#L138 not needed in this case

r0h4n: ahh, I missed that, you can add "include_current_node" as a flag for the original function and re-use it as you had suggested

tendrl/commons/flows/expand_cluster/__init__.py
r0h4n: the comment needs to be removed

tendrl/commons/flows/import_cluster/__init__.py
```diff
@@ -23,110 +23,111 @@ def run(self):
         integration_id = self.parameters['TendrlContext.integration_id']
```
r0h4n: We also need to ensure  this "integration_id"  does not exist at /clusters/$integration_id, if it does, raise FlowExecutionFailedError("Cluster (id) already exists!, cannot import another cluster with same id")

tendrl/commons/flows/expand_cluster/__init__.py
nnDarshan: ack

tendrl/commons/flows/expand_cluster/utils.py
nnDarshan: ack

tendrl/commons/flows/import_cluster/__init__.py
nnDarshan: ack

nnDarshan:
@r0h4n Please have a look. have incorporated your suggestions

r0h4n:
Looks good. Verified?

nnDarshan:
@r0h4n yes, Its verified

PR #452, opened 2017-05-10
Added better handling of ssh jobs during cluster CRUD.

nnDarshan:
This PR adds better logging of ssh jobs during cluster
creation/expansion/import so that suitable error message
is raised and The flow fails if any of tht job fails.
Also this adds a time limit of 10 mins for ssh jobs to
finisn, else its treated as failed.

tendrl-bug-id: https://github.com/Tendrl/node-agent/issues/422

### DIFF
--- a/tendrl/commons/flows/create_cluster/__init__.py
+++ b/tendrl/commons/flows/create_cluster/__init__.py
@@ -1,6 +1,7 @@
 # flake8: noqa
 
 import etcd
+import gevent
 import json
 import uuid
 
@@ -84,7 +85,7 @@ class CreateCluster(flows.BaseFlow):
                     tags += current_tags
                     NS.node_context.tags = list(set(tags))
                     NS.node_context.save()
-
+                gevent.sleep(3)
 
         # SSH setup jobs finished above, now install sds bits and create cluster
         if "ceph" in sds_name:
--- a/tendrl/commons/flows/expand_cluster/__init__.py
+++ b/tendrl/commons/flows/expand_cluster/__init__.py
@@ -1,4 +1,5 @@
 import etcd
+import gevent
 import json
 import uuid
 
@@ -57,6 +58,7 @@ class ExpandCluster(flows.BaseFlow):
                 )
 
                 all_ssh_jobs_done = True
+                gevent.sleep(3)
 
         # SSH setup jobs finished above, now install sds
         # bits and create cluster
--- a/tendrl/commons/flows/import_cluster/__init__.py
+++ b/tendrl/commons/flows/import_cluster/__init__.py
@@ -123,6 +123,7 @@ class ImportCluster(flows.BaseFlow):
                         # set gdeploy_provisioned to true so that no other nodes
                         # tries to configure gdeploy
                         self.parameters['gdeploy_provisioned'] = True
+                        gevent.sleep(3)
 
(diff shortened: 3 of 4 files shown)


nnDarshan:

tendrl/commons/flows/create_cluster/__init__.py
```diff
@@ -58,7 +59,8 @@ def run(self):
             )
 
         all_ssh_jobs_done = False
-        while not all_ssh_jobs_done:
+        count = 0
```
r0h4n: Please make improvements in detecting failure of below ssh authorize_key flow instead of defensively adding timers here

- https://github.com/Tendrl/commons/blob/develop/tendrl/commons/flows/authorize_ssh_key/__init__.py
In this flow check if the ssh key is actually present in the authorized_keys file and also try to ssh on the current node using the newly added ssh key

tendrl/commons/flows/expand_cluster/__init__.py
```diff
@@ -35,7 +36,8 @@ def run(self):
             )
 
         all_ssh_jobs_done = False
-        while not all_ssh_jobs_done:
+        count = 0
```
r0h4n: same as my comment above on authorize_keys flows

tendrl/commons/flows/import_cluster/__init__.py
```diff
@@ -94,7 +94,8 @@ def run(self):
                 )
 
                 all_ssh_jobs_done = False
-                while not all_ssh_jobs_done:
+                count = 0
```
r0h4n: same as my comment above on authorize_keys flows

tendrl/commons/flows/create_cluster/__init__.py
nnDarshan: I don't get, how this can be improved. As far as this flow is concerned, It has fired few jobs for writing authorized keys file targeted to other nodes, This flow has to make sure that all these jobs are finished to proceed further. If at all  any of these jobs are not completed after some time, We are clearly reporting the job ids of such jobs.
And in ssh key write job, we will clearly log errors if any that it encountered during writing to ssh keys file. If there are any errors, it means the authorized_key file is not updated, even If we check this we get the same result. All we can do is report the error that ansible encountered while writing the file.

So as far as end user is concerned, when he fires a create/import cluster job, if ssh_copy fails in any of the nodes. He will see a message saying "import failed because following ssh copy jobs failed: <list of ssh jobs that failed>". when he refers to any of those failed job-ids, he will get to know why exactly ssh copy failed.

r0h4n: I agree with you, but whats your take on  

https://github.com/Tendrl/commons/blob/develop/tendrl/commons/flows/authorize_ssh_key/__init__.py

"In this flow check if the ssh key is actually present in the authorized_keys file and also try to ssh on the current node using the newly added ssh key"

nnDarshan: I feel if there is any issue while copying, ansible will definitely return the error and we are raising exception with this error message.
I don't think we will land in a situation where this utility reports success and in reality the password-less ssh is not working(If at all such a situation arises, then we definitely will add the checks that you have mentioned)

nnDarshan:
@r0h4n Have removed timers as suggested

PR #453, opened 2017-05-10
WIP - Added flow for adding OSDs to existing OSD node

shtripat:

### DIFF
--- a/tendrl/commons/flows/create_cluster/ceph_help.py
+++ b/tendrl/commons/flows/create_cluster/ceph_help.py
@@ -1,5 +1,7 @@
 # flake8: noqa
+import etcd
 import gevent
+import json
 
 from tendrl.commons.event import Event
 from tendrl.commons.flows.exceptions import FlowExecutionFailedError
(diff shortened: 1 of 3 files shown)


tendrl/commons/objects/definition/master.yaml
```diff
@@ -71,6 +71,24 @@ namespace.tendrl:
       type: Create
       uuid: 2f94a48a-05d7-408c-b400-e27827f4edef
       version: 1
+    AddOsds:
```
r0h4n: This flow belongs to integrations.ceph.flows in the node agent (https://github.com/Tendrl/node-agent/tree/develop/tendrl/integrations/ceph)

Anything under the ceph_help files are actually intended to be integration specific flows under NS "integrations.$sds_name"  in node-agent

tendrl/commons/flows/create_cluster/ceph_help.py
```diff
@@ -161,6 +163,40 @@ def create_osds(parameters, created_mons):
...
+                        journal_details[journal_disk_name]['journal_count'] += 1
+                        journal_details[journal_disk_name]['ssd'] = True
+                    else:
+                        journal_details[journal_disk_name] = {
+                            'journal_count': 1,
+                            'ssd': False,
+                            'journal_size': config['journal_size'] * 1024 * 1024
+                        }
+
+            NS.etcd_orm.client.write(
```
r0h4n: - Avoid using the raw client writes, the reasons for this have been widely discussed in the past
- This is sds specific data and belongs to /clusters/$integration_id/ via a new object called integrations.ceph.objects.Journal (attrs: osd_uuid [0]... other journal attributes)
- also create integrations.ceph.objects.Osd, instances of which are created during ceph osd create.

[0]: http://ceph.com/geen-categorie/ceph-osd-uuid-conversion-to-osd-id-and-vice-versa/

tendrl/commons/objects/definition/master.yaml
shtripat: Ok. will move it there.

tendrl/commons/flows/create_cluster/ceph_help.py
shtripat: >> This is sds specific data and belongs to /clusters/$integration_id/ via a new object called
>> integrations.ceph.objects.Journal (attrs: osd_uuid 0... other journal attributes)

Will add a class for the same

>> also create integrations.ceph.objects.Osd, instances of which are created during ceph osd
>> create.

Do you mean to save OSD details separately as /clusters/{fsid}/osds other than maps?

r0h4n: > Do you mean to save OSD details separately as /clusters/{fsid}/osds other than maps?
Yes
btw it is not {fsid}, it's /clusters/:integration_id/Osds/:osd_uuid/attributes here... 


> Will add a class for the same. Also this data I need to save only for reference while journal mapping calculation to make sure SSDs are considered if not already 4 journals created on them. Actual OSD jorunal details would be available under maps with partition path etc.

Create the integrations.ceph.objects.Journal  and include journal detail from maps and also store the data that you require in this PR

shtripat: May be as part of another PR?

shtripat:
Closing as add osd flow is moved to node-agent integrations

PR #466, opened 2017-05-13
Error handling for boundary cases

rishubhjain:
There might me cases when node-agent is not configured/ non-active and hence might not have object keys such as config, NodeContext or TendrlContext so the setup_common_objects might throw error, I feel there is a need for error handling

@r0h4n  is the presence of node context necessary or can it be surrounded by try except?


r0h4n:
This PR is not required,  we already check if the object in doubt is present in the current_ns

https://github.com/Tendrl/commons/blob/develop/tendrl/commons/__init__.py#L379

PR #492, opened 2017-05-17
New disk structure for for store disk details

GowthamShanmugam:

### DIFF
--- a/tendrl/commons/objects/block_device/__init__.py
+++ b/tendrl/commons/objects/block_device/__init__.py
@@ -1,0 +1,58 @@
+from tendrl.commons import objects
+
+
+class BlockDevice(objects.BaseObject):
+    def __init__(self, device_name=None, device_kernel_name=None,
+                 parent_name=None, major_to_minor_no=None, fstype=None,
+                 mount_point=None, label=None, fsuuid=None,
+                 read_ahead=None, read_only=None, removable_device=None,
+                 size=None, state=None, owner=None, group=None,
+                 mode=None, alignment=None, min_io_size=None,
+                 optimal_io_size=None, phy_sector_size=None,
+                 log_sector_size=None, device_type=None,
+                 scheduler_name=None, req_queue_size=None,
+                 discard_align_offset=None, discard_granularity=None,
+                 discard_max_bytes=None, discard_zeros_data=None,
+                 used=None, rotational=None, ssd=None,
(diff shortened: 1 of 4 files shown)


GowthamShanmugam:
@r0h4n @brainfunked @anivargi @nnDarshan @nthomas-redhat @shtripat please review

r0h4n:
Verified?

brainfunked:
Given the different types of objects coming in for the local storage, I think we should move everything (including the block devices hierarchy in https://github.com/Tendrl/node-agent/pull/443) under a "local storage" directory per node.

This would allow the following structure to take shape:

* local_storage (or LocalStorage, depending upon the implementation)
  * Disks: contains only physical disks (HDD, SSD, NVME) attached to the node and their hardware and partitioning information.
  * Virtio: virtio devices don't show up as physical disks with hardware properties on a virtual machine. It would thus be better to separate them out into their own directory. This allows Disks to contain only physical drives.
  * BlockDevices: contains all the block devices created by the system to access various local storage devices.
  * RawReferences: mapping of block devices using various attributes generated by the system under the /dev/disk hierarchy.

In the future, additional directories for LVM, RAID etc. could be added, as required. In addition, any kind of mapping between a block device file (vda, sda, md0 etc.) to the corresponding hardware (or software) device could be added in the local_storage directory itself, without polluting any of the specific hierarchies.

GowthamShanmugam:
@brainfunked 
 physical devices are stored like : nodes/{node_id}/LocalStorage/Disks/vendor_device_serialno
virtual devices are stored like:  nodes/{node_id}/LocalStorage/Virtio/by-id
block devices are stored like:   nodes/{node_id}/LocalStorage/BlockDevices/all/device_name
                                                 nodes/{node_id}/LocalStorage/BlockDevices/used/device_name
                                                 nodes/{node_id}/LocalStorage/BlockDevices/free/device_name
raw_references are stored like:  nodes/{node_id}/LocalStorage/RawReferences

is this correct?

r0h4n:
@brainfunked Please check out @GowthamShanmugam question above

brainfunked:
@GowthamShanmugam Correct. However, whether we use by-id or by-uuid for virtio devices would be determined based on the conversation at https://github.com/Tendrl/node-agent/issues/463. In any case, I'll sum the discussions up at a spec issue.

PR #510, opened 2017-05-19
Fixed issue while checking the status of SSH jobs for nodes

shtripat:
It used to break from loop in first iteration and fixed the same.

tendrl-bug-id: Tendrl/node-agent#451


shtripat:
Thanks @GowthamShanmugam for debugging and providing the solution.
Verified with fresh set of nodes and create + import is successful.

PR #521, opened 2017-05-19
Fix for getting disk id while parsing hwinfo output

GowthamShanmugam:


GowthamShanmugam:


here i have solved some syntax error in some other file also

PR #529, opened 2017-05-23
Make logging_socket_path, node-id and publisher optional

anmolbabu:

### DIFF
--- a/tendrl/commons/message.py
+++ b/tendrl/commons/message.py
@@ -50,6 +50,8 @@ class Message(object):
         else:
             self.caller = caller
         self.priority = priority
+        if not publisher:
+            publisher = NS.publisher_id
         self.publisher = publisher
         self.node_id = node_id
         if self.node_id is None:
--- a/tendrl/commons/utils/ansible_module_runner.py
+++ b/tendrl/commons/utils/ansible_module_runner.py
@@ -31,26 +31,46 @@ class AnsibleRunner(object):
 
     """
 
-    def __init__(self, module_path, **kwargs):
+    def __init__(
+        self,
+        module_path,
+        publisher=None,
+        node_id=None,
+        socket_path=None,
+        **kwargs
+    ):
         self.module_path = modules.__path__[0] + "/" + module_path
+        self.publisher = publisher
+        self.node_id = node_id
+        self.socket_path = socket_path
+        if not publisher:
+            self.publisher = NS.publisher_id
+        if not node_id:
+            self.node_id = NS.node_context.node_id
+        if not socket_path:
+            self.socket_path = NS.config.data['logging_socket_path']
         if not os.path.isfile(self.module_path):
             Event(
                 Message(
                     priority="debug",
-                    publisher=NS.publisher_id,
+                    publisher=self.publisher,
                     payload={"message": "Module path: %s does not exist" %
                                         self.module_path
-                             }
-                )
+                             },
+                    node_id=self.node_id
+                ),
+                socket_path=self.socket_path
             )
             raise AnsibleModuleNotFound(module_path=self.module_path)
         if kwargs == {}:
             Event(
                 Message(
                     priority="error",
-                    publisher=NS.publisher_id,
-                    payload={"message": "Empty argument dictionary"}
-                )
+                    publisher=self.publisher,
+                    payload={"message": "Empty argument dictionary"},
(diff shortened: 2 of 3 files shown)


anmolbabu:
Primary Verification Done but since this is used by multiple places in code I would like to verify it thoroughly....

tendrl/commons/message.py
```diff
@@ -50,6 +50,8 @@ def __init__(self, priority, publisher, payload, job_id=None,
         else:
             self.caller = caller
         self.priority = priority
+        if not publisher:
+            publisher = NS.publisher_id
```
rishubhjain: what if the NS.publisher_id returns None, we have handled such a case here https://github.com/Tendrl/commons/blob/develop/tendrl/commons/message.py#L94,
we pass publisher like https://github.com/Tendrl/commons/blob/develop/tendrl/commons/__init__.py#L32

rishubhjain:
For logging an Event with message, there is this function https://github.com/Tendrl/commons/blob/develop/tendrl/commons/utils/log_utils.py which saves you trouble  of handling event and message errors. It could be used like https://github.com/Tendrl/commons/blob/develop/tendrl/commons/__init__.py#L32

r0h4n:
@anmolbabu Is the ansible module runner used somewhere in your collectd plugins? And is that the reason you'd like to make the publisher_id optional ?

anmolbabu:
@r0h4n Yes but not directly. Collectd plugin uses brick_utilization plugin in gluster-integration which internally uses AnsibleModuleRunner which is logging success and/or failure with severities accordingly. So this is required.... Also the collectd plugin uses cmd utils which also utilizes the ansible wrapper.

r0h4n:
@anmolbabu thanks, please make use of https://github.com/Tendrl/commons/blob/develop/tendrl/commons/utils/log_utils.py 

example: https://github.com/Tendrl/commons/blob/develop/tendrl/commons/__init__.py#L32

You can make changes to the log_util if required

tendrl/commons/message.py
anmolbabu: So what is the problem here?
If its already there it means its already checked right?
So, if NS.publisher_id is none then when the message object is passed to Event class, it is validated there as in :
https://github.com/Tendrl/commons/blob/develop/tendrl/commons/event.py#L17->
https://github.com/Tendrl/commons/blob/develop/tendrl/commons/message.py#L69

You are probably trying to say wherever/whoever is trying to use Message and/or Event class should only go through log_utils just so that:
1. These validations are performed as the order that ensures this validation -> Event class creation accompanying Message class creation and ensuring no existence of object of Message class without it being passed to Event class -- I am not convinced of whether this is the correct way...
2. If you are saying that using log_utils, having to make 2 calls (1 to create Message object and 1 to create Event object) can be avoided, this is probably not the correct place to comment. And the scope of this PR is not that.
There probably needs to be a separate PR to do that mass change.

anmolbabu:
@r0h4n @rishubhjain The scope of this PR was to make certain fields optional and no new messages were added. I am wary of introducing any regressions by making such a change at this point....

In my opinion this should be done as part of a separate PR on each component ... so that in worst case reverting the change becomes easier if required...

tendrl/commons/message.py
rishubhjain: I guess the issue would be when publisher_id itself is not present in NS, it will throw a KeyError which is not handled here, I suppose if what you suggested should be implemented then either we will have to handle this error case or make sure no one calls this function until NS.publisher_id is atleast present i.e NS has publisher _id key in its dict

rishubhjain: @anmolbabu I guess this error case must be considered

anmolbabu: @rishubhjain Why can't the key error check be added to validate function?

rishubhjain: @anmolbabu it can be may be, according to me each function must not worry about all the errors , some information should be compulsory for the function to work else  everything will have to be placed under try and except, but I am not sure if this is the right way to code, and it's a small change 

@r0h4n if possible could you give your opinion , I can make the amends as per suggestions

anmolbabu: Well I see no problem in a class member function validating values of all the class data members specially when you say a utility outside the class can do it....

anmolbabu:
Not required anymore

PR #534, opened 2017-05-23
Added patch for minimum required gluster version

shtripat:
Bug-URl: https://bugzilla.redhat.com/1453102


tendrl/commons/0001-Set-the-minimum-required-gluster-version-for-downstr.patch
```diff
@@ -0,0 +1,27 @@
...
+diff --git a/tendrl/commons/objects/definition/master.yaml b/tendrl/commons/objects/definition/master.yaml
+index 9fffc18..bec33d4 100644
+--- a/tendrl/commons/objects/definition/master.yaml
++++ b/tendrl/commons/objects/definition/master.yaml
+@@ -3,7 +3,7 @@ namespace.tendrl:
+   supported_sds:
+    - ceph
+    - gluster
+-  min_reqd_gluster_ver: 3.9.0
++  min_reqd_gluster_ver: 3.8.4
```
r0h4n: You sure 3.8.4 is the earliest gluster version with "get-state"  cli? 

@mkudlej please confirm if 3.8.4 is the earliest//oldest downstream gluster version with "get-state" cli support

tendrl/commons/0001-Set-the-minimum-required-gluster-version-for-downstr.patch
```diff
@@ -0,0 +1,27 @@
+From 6b509b919b480f3207d8bd789413db66ce68711d Mon Sep 17 00:00:00 2001
```
r0h4n: Please save this patch at etc/patches/0001-minimum-rhgs-version-rhel7.patch

tendrl/commons/0001-Set-the-minimum-required-gluster-version-for-downstr.patch
atinmu: If the check is the gluster bit version then yes this should be correct. In downstream rhgs-3.2 since we rebased to 3.8.4 'gluster --version' will have it. In upstream this is available in 3.9

shtripat:
Sure Rohan. Will update the PR...

etc/patches/0001-minimum-rhgs-version-rhel7.patch
```diff
@@ -0,0 +1,27 @@
...
+diff --git a/tendrl/commons/objects/definition/master.yaml b/tendrl/commons/objects/definition/master.yaml
+index 9fffc18..bec33d4 100644
+--- a/tendrl/commons/objects/definition/master.yaml
++++ b/tendrl/commons/objects/definition/master.yaml
+@@ -3,7 +3,7 @@ namespace.tendrl:
+   supported_sds:
+    - ceph
+    - gluster
+-  min_reqd_gluster_ver: 3.9.0
++  min_reqd_gluster_ver: 3.8.4
```
mkudlej: But this is not true in upstream, right? So I would like to suggest:

1. have check here for all required features or
2. have this simple patch only for downstream version

shtripat: As discussed with @r0h4n this .patch would be applied only while downstream builds and upstream builds would happen with default file under `objects/definition/master.yaml`.

@r0h4n can you confirm this?

mkudlej: @shtripat OK, if this patch is applied only in DS specfile I agree with that.

shtripat:
>> Is the downstream spec file updated to apply this patch, and is that tested?

@TimothyAsir we need to apply this patch and build downstream rpms. We would need spec changes to apply the patch. Once done I can install and verify the changes may be.

shtripat:
>> Is the downstream spec file updated to apply this patch, and is that tested?

I modified tendrl-commons.spec locally to add patch file and built the rpms using rpmbuild. I installed it and I see the correct yaml file with minimum_gluster_version as 3.8.4 populated.

I feel this much should be fine.

@TimothyAsir you need to test once with downstream spec file

PR #567, opened 2017-06-01
Corrected etcd dir path for BlockDevices and Disks

shtripat:

### DIFF
--- a/tendrl/commons/objects/node/atoms/is_node_tendrl_managed/__init__.py
+++ b/tendrl/commons/objects/node/atoms/is_node_tendrl_managed/__init__.py
@@ -50,31 +50,6 @@ class IsNodeTendrlManaged(objects.BaseAtom):
                     "Node doesnt have Memory details populated"
                 )
 
-            # Check if node has the block devices populated
-            try:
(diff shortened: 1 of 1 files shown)


tendrl/commons/objects/node/atoms/is_node_tendrl_managed/__init__.py
```diff
@@ -52,7 +52,7 @@ def run(self):
 
             # Check if node has the block devices populated
             try:
-                block_devices = NS._int.client.read("nodes/%s/BlockDevices" % node_id)
+                block_devices = NS._int.client.read("nodes/%s/LocalStorage/BlockDevices" % node_id)
```
r0h4n: shouldnt we be checking "nodes/%s/LocalStorage/BlockDevices/all"

r0h4n: and you missed LocalStorage/virtio 

We should not be doing these disk checks at all, because we cannot be sure if its going to show up as virtio or blockdevice or disk

I am removing the disk checks for now

PR #580, opened 2017-06-06
Make logging_socket_path, node_id and publisher_id optional

anmolbabu:
1. tendrl's threshold notification handling collectd plugin
   uses the tendrl commons provided service utility to restart
   collectd after config is generated.
2. service utility uses ansible_module_runner for actions on service
Now the above said collectd plugin being an external integration, doesn't
have a namespace of its own and hence this pr facilitates optionally passing
logging_socket_path, node_id and publisher_id explicitly so that the
Message-Event framework can use these plugin passed values instead of
accessing them from NS which does not exist.

### DIFF
--- a/tendrl/commons/utils/ansible_module_runner.py
+++ b/tendrl/commons/utils/ansible_module_runner.py
@@ -31,26 +31,40 @@ class AnsibleRunner(object):
 
     """
 
-    def __init__(self, module_path, **kwargs):
+    def __init__(
+        self,
+        module_path,
+        publisher_id=None,
+        node_id=None,
+        socket_path=None,
+        **kwargs
+    ):
         self.module_path = modules.__path__[0] + "/" + module_path
+        self.socket_path = socket_path or NS.config.data['logging_socket_path']
+        self.publisher_id = publisher_id or NS.publisher_id
+        self.node_id = node_id or NS.node_context.node_id
         if not os.path.isfile(self.module_path):
             Event(
                 Message(
                     priority="debug",
-                    publisher=NS.publisher_id,
+                    publisher=self.publisher_id,
                     payload={"message": "Module path: %s does not exist" %
                                         self.module_path
-                             }
-                )
+                             },
+                    node_id=self.node_id
+                ),
+                socket_path=self.socket_path
             )
             raise AnsibleModuleNotFound(module_path=self.module_path)
         if kwargs == {}:
             Event(
                 Message(
                     priority="debug",
-                    publisher=NS.publisher_id,
-                    payload={"message": "Empty argument dictionary"}
-                )
+                    publisher=self.publisher_id,
+                    payload={"message": "Empty argument dictionary"},
+                    node_id=self.node_id
+                ),
+                socket_path=self.socket_path
             )
             raise ValueError
         else:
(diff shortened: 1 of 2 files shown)


anmolbabu:
Initial verification Done...
Probably need to verify atleast few minor flows that make use of this framework

anmolbabu:

tendrl/commons/utils/ansible_module_runner.py
```diff
@@ -31,26 +31,42 @@ class AnsibleRunner(object):
...
+        node_id=None,
+        socket_path=None,
+        **kwargs
+    ):
         self.module_path = modules.__path__[0] + "/" + module_path
+        self.socket_path = socket_path
+        self.publisher_id = publisher_id
+        if not self.publisher_id:
+            self.publisher_id = NS.publisher_id
+        self.node_id = node_id
```
r0h4n: what if node_id is None?

Use
self.node_id = node_id or NS.node_context.node_id

anmolbabu:
The epicenter of failure was at https://github.com/Tendrl/commons/pull/580/files#diff-52a44eabbad36a1b132c91c25e587e1bR52 on rhel setup with ansible-2.2.3.0-1.el7.noarch
and never felt on centos setup with ansible-2.3.0.0-3.el7.noarch

tendrl/commons/utils/ansible_module_runner.py
```diff
@@ -31,26 +31,42 @@ class AnsibleRunner(object):
...
+        self,
+        module_path,
+        publisher_id=None,
+        node_id=None,
+        socket_path=None,
+        **kwargs
+    ):
         self.module_path = modules.__path__[0] + "/" + module_path
+        self.socket_path = socket_path
+        self.publisher_id = publisher_id
```
r0h4n: Use

self.publisher_id = publisher_id or NS.publisher_id

tendrl/commons/utils/ansible_module_runner.py
```diff
@@ -31,26 +31,42 @@ class AnsibleRunner(object):
...
+    def __init__(
+        self,
+        module_path,
+        publisher_id=None,
+        node_id=None,
+        socket_path=None,
+        **kwargs
+    ):
         self.module_path = modules.__path__[0] + "/" + module_path
+        self.socket_path = socket_path
```
r0h4n: use
self.socket_path = socket_path or "/var/run/tendrl/message.sock"

tendrl/commons/utils/ansible_module_runner.py
anmolbabu: They are handled in Message-Event combo as in https://github.com/Tendrl/commons/blob/develop/tendrl/commons/message.py#L55 and https://github.com/Tendrl/commons/blob/develop/tendrl/commons/event.py#L28

tendrl/commons/utils/ansible_module_runner.py
anmolbabu: They are handled in Message-Event combo as in https://github.com/Tendrl/commons/blob/develop/tendrl/commons/message.py#L55 and https://github.com/Tendrl/commons/blob/develop/tendrl/commons/event.py#L28

tendrl/commons/utils/ansible_module_runner.py
anmolbabu: Done

anmolbabu:
Verified

PR #573, opened 2017-06-03
Fixing Issue #571,#582 and #572

rishubhjain:

PR #581, opened 2017-06-06
Mark LocalStorage/Disk/:disk_id/ssd

r0h4n:
Tendrl marks the BlockDevices with ssd flag
Also required to mark raw disks with ssd flag

tendrl-bug-id: Tendrl/node-agent#489


rishubhjain:
@r0h4n There was a ", " missing in virtual_disk/__init__.py 
https://github.com/Tendrl/commons/blob/develop/tendrl/commons/objects/virtual_disk/__init__.py#L22
I have added its patch with https://github.com/Tendrl/commons/pull/573

PR #593, opened 2017-06-14
Added logic to enable integration services while import flow

shtripat:
The services should be enabled to make sure that across node
restarts these services get started again.


shtripat:

shtripat:
Verified the changes with ceph cluster import. Need to verify once with create flow as well.

r0h4n:
File an issue

shtripat:
Closing this as it would be taken care as part of spec/service file changes.

Issue #599, opened 2017-06-16
Ceph cluster create  job is not updated in case of failure


nthomas-redhat:
Cluster create job stuck in processing in case of any failure returned by ceph-installer

nthomas-redhat:
Issue is with job update where dict is passed string is expexted;


Jun 16 08:38:26 dhcp42-247 journal: 2017-06-16 03:08:26.364186+00:00 - node_agent - /usr/lib/python2.7/site-packages/tendrl/commons/jobs/__init__.py:271 - _run - ERROR - Job (f4c20d6d-da82-4506-9c35-9074d2e7cafe) processing failure, error:coercing to Unicode: need string or buffer, dict found - TypeError: [{u'function': u'run', u'line': 327, u'file': u'/usr/lib64/python2.7/site-packages/gevent/greenlet.py', u'statement': u'result = self._run(*self.args, **self.kwargs)'}, {u'function': u'_run', u'line': 235, u'file': u'/usr/lib/python2.7/site-packages/tendrl/commons/jobs/__init__.py', u'statement': u'payload={"message": _msg + e.message,'}]
Jun 16 08:38:26 dhcp42-247 tendrl-node-agent: 2017-06-16 03:08:26.364186+00:00 - node_agent - /usr/lib/python2.7/site-packages/tendrl/commons/jobs/__init__.py:271 - _run - ERROR - Job (f4c20d6d-da82-4506-9c35-9074d2e7cafe) processing failure, error:coercing to Unicode: need string or buffer, dict found - TypeError: [{u'function': u'run', u'line': 327, u'file': u'/usr/lib64/python2.7/site-packages/gevent/greenlet.py', u'statement': u'result = self._run(*self.args, **self.kwargs)'}, {u'function': u'_run', u'line': 235, u'file': u'/usr/lib/python2.7/site-packages/tendrl/commons/jobs/__init__.py', u'statement': u'payload={"message": _msg + e.message,'}]

PR #600, opened 2017-06-16
Ceph cluster create job is not updated in case of failure

nthomas-redhat:
tendrl-bug-id: Tendrl/commons#599

Signed-off-by: Nishanth Thomas <nthomas@redhat.com>


tendrl/commons/flows/create_cluster/ceph_help.py
```diff
@@ -222,12 +222,10 @@ def wait_for_task(task_id):
                 if resp["succeeded"]:
                     return
                 else:
-                    stderr = resp.get("stderr", "ceph-installer task_id %s failed and did not complete" % task_id)
-                    stdout = resp.get("stdout", "")
-                    raise FlowExecutionFailedError(dict(ceph_installer_task_id=task_id, ceph_installer_task_stdout=stdout,
-                                        ceph_installer_task_stderr=stderr))
+                    stderr = resp.get("stderr", "")
+                    raise FlowExecutionFailedError("ceph-installer task_id %s \
```
shtripat: Exception was being raised earlier also. How come it was not working then?

r0h4n:
If there's a failure in the job consumer thread, we need to fix it there. Ill send a fix, please test with that.

tendrl/commons/flows/create_cluster/ceph_help.py
shtripat: Ah got the answer looking at the issue description. I am ok with this change.

May be small suggestion

```
raise FlowExecutionFailedError(
    "ceph-installer task_id %s failed and didnt complete. Error: %s" %
    (task_id, stderr)
)
```

r0h4n:
Closing this PR, this is not really a fix of the original problem described here https://github.com/Tendrl/commons/issues/599#issuecomment-308930145

Please use below fix and test:
https://github.com/Tendrl/commons/commit/3c477c1c33c297cf7f8e76eb94e12b3131831a43

PR #618, opened 2017-06-19
Exposed sds sync interval as configuration

shtripat:
tendrl-bug-id: Tendrl/ceph-integration#299
tendrl-bug-id: Tendrl/gluster-integration#312


shtripat:

PR #619, opened 2017-06-19
Update tendrl-commons.spec

rishubhjain:
https://github.com/Tendrl/commons/blob/develop/tendrl/commons/profiler.py#L12 requires GreenletProfiler which I guess is not installed with already existing pre-requisites.


r0h4n:
Profiling is optional : https://github.com/Tendrl/documentation/wiki/Tendrl-profiling

Please discuss such topics via a new issue before sending PR

Issue #585, opened 2017-06-08
Tendrl BaseObject needs a load_all() 


r0h4n:
Tendrl BaseObject should provide a load_all() function which loads all instances of the object from central store

- use Object's tendrl definition "value" to extract location where the list of primary keys can be found for that object.

r0h4n:
@rishubhjain please take a look at this issue, ask here if any questions

rishubhjain:
@r0h4n As per my understanding, the already existing load function in BaseObject loads the object of the  class (inheriting BaseObject), and the load_all() function must load all the instances of this class(inheriting BaseObject), am I correct?

r0h4n:
BaseObject.load() function loads that instance of the class from etcd,

BaseObject.load_all() should load all instances of the class from etcd

makes sense?

rishubhjain:
Yes,
The class mentioned here is referring to the class that is inheriting BaseObject Metaclass right?

rishubhjain:
@r0h4n I guess  we can close this issue.

PR #621, opened 2017-06-19
Loading all instances of object (Issue #585)

rishubhjain:
Load_all() function to load all instances of object
Returns List of instances that is loaded
tendrl-bug-id: Tendrl/commons#585

### DIFF
--- a/tendrl/commons/objects/__init__.py
+++ b/tendrl/commons/objects/__init__.py
@@ -162,6 +162,23 @@ class BaseObject(object):
                 NS._int.wreconnect()
                 NS._int.wclient.write(item['key'], item['value'], quorum=True)
 
+    def load_all(self):
+        value = '/'.join(self.value.split('/')[:-1])
+        try:
+            etcd_resp = NS._int.client.read(value)
+        except (etcd.EtcdConnectionFailed, etcd.EtcdException) as ex:
+                    if type(ex) != etcd.EtcdKeyNotFound:
+                        NS._int.reconnect()
+                        etcd_resp = NS._int.client.read(value)
+                    else:
+                        return None
+        ins = []
+        for item in etcd_resp.leaves:
+            self.value = item.key
+            ins.append(self.load())
+        return ins
+
+
     def load(self):
         if not "Message" in self.__class__.__name__:
             try:
--- a/tendrl/commons/tests/objects/test_objects_init.py
+++ b/tendrl/commons/tests/objects/test_objects_init.py
@@ -9,6 +9,8 @@ import mock
 from tendrl.commons import objects
 from tendrl.commons import TendrlNS
 from mock import patch
+import importlib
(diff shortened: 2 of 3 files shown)


rishubhjain:
@r0h4n Please review

tendrl/commons/objects/__init__.py
```diff
@@ -162,6 +162,16 @@ def save(self, update=True, ttl=None):
                 NS._int.wreconnect()
                 NS._int.wclient.write(item['key'], item['value'], quorum=True)
 
+    def load_all(self):
+        value = '/'.join(self.value.split('/')[:-1])
+        etcd_resp = NS._int.client.read(value)
+        ins = []
+        for item in etcd_resp.leaves:
+            self.value = item.key
+            ins.append(self.load())
```
GowthamShanmugam: have you tested this list, is it giving different instance

rishubhjain: yes I have tested it :)

tendrl/commons/objects/__init__.py
```diff
@@ -162,6 +162,16 @@ def save(self, update=True, ttl=None):
                 NS._int.wreconnect()
                 NS._int.wclient.write(item['key'], item['value'], quorum=True)
 
+    def load_all(self):
+        value = '/'.join(self.value.split('/')[:-1])
+        etcd_resp = NS._int.client.read(value)
```
anmolbabu: You might need to handle EtcdConnectionFailed as in https://github.com/Tendrl/commons/pull/621/files/6568a8bd1357aef13a3411c4a43577fdc8a9985e#diff-fbc8ea2e973a12d3591504a5230f8186R186

rishubhjain: @anmolbabu Thanks for the review ,I have made the changes.
I feel the wrapper would generalise the error handling although almost everytime the error is handled in the same/similar way

anmolbabu: @rishubhjain Yes and that generalization is what is required not just because it avoids having to do certain exception handling steps but because any read from etcd should have that resistance(for the moment, it just does a reconnect(creates new etcd client connection) when the connection fails but whenever we add any extra resilience the new channelized approach lets every access to etcd benefit from that resilience) built..

rishubhjain: @anmolbabu True, I will create a wrapper in tendrl/commons/utils which will have read/write function  that can be called and will handle generalised exceptions

rishubhjain: @anmolbabu I have filed an issue https://github.com/Tendrl/commons/issues/624 , will create the wrapper and submit a PR

rishubhjain:
@r0h4n thanks for the review. should I add the unit test case with this PR?

r0h4n:
@rishubhjain Yep

PR #622, opened 2017-06-19
Check pre-existing provisioner while importing a gluster cluster

shtripat:
If there is already a node marked as gluster provisioner with
tag `provisioner/gluster` mark the same as provisioner while
import cluster flow.

tendrl-bug-id: Tendrl/gluster-integration#315

### DIFF
--- a/tendrl/commons/flows/import_cluster/__init__.py
+++ b/tendrl/commons/flows/import_cluster/__init__.py
@@ -1,3 +1,4 @@
+import json
 import uuid
 
 import etcd
@@ -15,6 +16,27 @@ from tendrl.commons.message import ExceptionMessage
 
 
 class ImportCluster(flows.BaseFlow):
+    def _probe_and_mark_provisioner(self, node_list, integration_id):
+        try:
+            gluster_prov_nodes = json.loads(
+                NS._int.client.read("indexes/tags/provisioner/gluster").value
+            )
+            if gluster_prov_nodes and gluster_prov_nodes[0] in node_list:
+                # Ideally only one node would be present in this list
+                # Even if more than one in the list, mark one of them
+                # with tag `provisioner/{integration-id}` for further
+                # cluster operations
+                nc = NS.tendrl.objects.NodeContext(
+                    node_id=gluster_prov_nodes[0]
+                ).load()
+                tags = nc.tags
+                tags += ["provisioner/%s" % integration_id]
+                nc.tags = list(set(tags))
+                nc.save()
+                return True
+        except etcd.EtcdKyNotFound:
+            return False
+
     def run(self):
         try:
             # Lock nodes
@@ -74,7 +96,9 @@ class ImportCluster(flows.BaseFlow):
 
                 # check if gdeploy in already provisioned in this cluster
                 # if no it has to be provisioned here
-                if not self.parameters.get('gdeploy_provisioned', False) and sds_name.find("gluster") > -1:
+                if sds_name.find("gluster") > -1 and \
+                    not self.parameters.get("gdeploy_provisioned", False) and \
+                    not self._probe_and_mark_provisioner(self.parameters["Node[]"], integration_id):
                     create_cluster_utils.install_gdeploy()
                     create_cluster_utils.install_python_gdeploy()
                     ssh_job_ids = create_cluster_utils.gluster_create_ssh_setup_jobs(
--- a/tendrl/commons/objects/node_context/__init__.py
+++ b/tendrl/commons/objects/node_context/__init__.py
@@ -125,5 +125,5 @@ class NodeContext(objects.BaseObject):
             return None
 
     def render(self):
-        self.value = self.value.format(NS.node_context.node_id)
+        self.value = self.value.format(self.node_id or NS.node_context.node_id)
         return super(NodeContext, self).render()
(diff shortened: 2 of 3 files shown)


shtripat:
This is not yet verified

tendrl/commons/flows/import_cluster/__init__.py
```diff
@@ -15,6 +16,30 @@
 
 
 class ImportCluster(flows.BaseFlow):
+    def _probe_and_mark_provisioner(self, node_list, integration_id):
+        for node in node_list:
+            try:
+                tags = json.loads(
+                    NS._int.client.read(
+                        "nodes/%s/NodeContext/tags" % node
+                    ).value
```
r0h4n: Create NS.tendrl.objects.NodeContext(node_id=node) instead of all these raw read writes

tendrl/commons/flows/import_cluster/__init__.py
```diff
@@ -15,6 +16,30 @@
...
+                )
+                if "provisioner/gluster" in tags and \
+                    "provisioner/%s" % integration_id not in tags:
+                    tags.append("provisioner/%s" % integration_id)
+                    NS._int.wclient.write(
```
r0h4n: NodeContext.tags is a "list" type in Tendrl definitions, which means it is supposed to be a json array. Hence this is going to fail.

Create an actual NS.tendrl.objects.NodeContext(node_id=node) and add the new tag to its "tags" attribute and call .save()

tendrl/commons/flows/import_cluster/__init__.py
shtripat: Tried this but doesnt work as node context class's render() function doesnt consider the passed node_id and always uses current node's id only 

```
    def render(self):
        self.value = self.value.format(NS.node_context.node_id)
        return super(NodeContext, self).render()
```

If we change that, yes this will work.

tendrl/commons/flows/import_cluster/__init__.py
shtripat: as above

shtripat:
Verified the below flows
1. created a cluster from tendrl UI
2. cleaned up the etcd
3. imported the cluster from UI (it marked the existing provisioner node with `provisioner/{integration-id}`)
4. created bricks in the imported cluster
5. created volume using the created bricks in step-4

Works as expected.

tendrl/commons/flows/import_cluster/__init__.py
r0h4n: Make the change in node context render

shtripat: Done in latest push below.

tendrl/commons/flows/import_cluster/__init__.py
```diff
@@ -15,6 +15,25 @@
 
 
 class ImportCluster(flows.BaseFlow):
+    def _probe_and_mark_provisioner(self, node_list, integration_id):
+        for node in node_list:
```
r0h4n: Instead of iterating over list of nodes and loading every node's NodeContext

1) fetch the list of node for tag "provisioner/gluster"  from `/indexes/tags/provisioner/gluster`

The from set(node_list) - set(list_of_node_from_step_1) , you can figure out the node with "provisioner/gluster" tag

tendrl/commons/objects/node_context/__init__.py
```diff
@@ -125,5 +125,5 @@ def _get_node_id(self):
             return None
 
     def render(self):
-        self.value = self.value.format(NS.node_context.node_id)
+        self.value = self.value.format(self.node_id)
```
shtripat: Done

tendrl/commons/flows/import_cluster/__init__.py
shtripat: Done. will try this.

shtripat:
Verified the same scenarios mentioned in https://github.com/Tendrl/commons/pull/622#issuecomment-309686058. Works fine.

tendrl/commons/flows/import_cluster/__init__.py
```diff
@@ -15,6 +16,27 @@
 
 
 class ImportCluster(flows.BaseFlow):
+    def _probe_and_mark_provisioner(self, node_list, integration_id):
+        try:
+            gluster_prov_nodes = json.loads(
+                NS._int.client.read("indexes/tags/provisioner/gluster").value
+            )
+            if gluster_prov_nodes and len(gluster_prov_nodes) > 0:
```
r0h4n: len check not required

tendrl/commons/flows/import_cluster/__init__.py
```diff
@@ -15,6 +16,27 @@
...
+                # Ideally only one node would be present in this list
+                # Even if more than one in the list, mark one of them
+                # with tag `provisioner/{integration-id}` for further
+                # cluster operations
+                nc = NS.tendrl.objects.NodeContext(
+                    node_id=gluster_prov_nodes[0]
```
r0h4n: I had requested to check if gluster_prov_nodes[0] is in node_list, this current code can also add a completely different cluster node , so make that check

tendrl/commons/flows/import_cluster/__init__.py
shtripat: Done

tendrl/commons/flows/import_cluster/__init__.py
shtripat: Done

Issue #624, opened 2017-06-20
Wrapper for read/write to etcd


rishubhjain:
As per discussion with @anmolbabu https://github.com/Tendrl/commons/pull/621#discussion_r122975898 I am writing a wrapper function for reading and writing data in etcd with proper exception handling. The generalised approach will be useful to provide right error handling to all access to etcd through out the code.

PR #625, opened 2017-06-20
Wrapper for read/write to etcd

rishubhjain:
tendrl-bug-id: Tendrl/commons#624

### DIFF
--- a/tendrl/commons/utils/etcd_utils.py
+++ b/tendrl/commons/utils/etcd_utils.py
@@ -1,0 +1,68 @@
+import etcd
+import __builtin__
+
+'''
+   Read from etcd
+   input params:
+       key  : type  - >  string
+              value - >  attr to be fetched
+
+   return param:
+       dict - > if read is successful
+       None - > if key is not found
+'''
+ 
+def read(key):
+    try:
+        return NS._int.client.read(key)
+    except (etcd.EtcdConnectionFailed, etcd.EtcdException) as ex:
+        if type(ex) != etcd.EtcdKeyNotFound:
+            NS._int.reconnect()
+            return NS._int.client.read(key)
+        else:
+            return None
+
+
+'''
+   Write to etcd
+   input params:
+       key   : type  - >  string
+               value - >  attr to be inserted
+       value : type  - >  string/int etc
+               value - >  value of attr to be inserted
+       quorum: type  - >  bool
+               value - >  value for quorum (True/False)
+                          default : True
+
+   return param:
+       None
+'''
+
+def write(key,value,quorum = True):
+    try:
+        NS._int.wclient.write(key,value,quorum = quorum)
+    except (etcd.EtcdConnectionFailed, etcd.EtcdException) as ex:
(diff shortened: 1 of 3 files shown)


tendrl/commons/utils/etcd_utils.py
```diff
@@ -0,0 +1,73 @@
...
+              value - >  Ns._int
+       key  : type  - >  string
+              value - >  attr to be fetched
+
+   return param:
+       dict - > if read is successful
+       None - > if key is not found
+'''
+
+def read(_int,key):
```
anmolbabu: Any specific reasons for having the _int as a parameter??? I feel that's not required @r0h4n can  comment better though

anmolbabu: That should be NS._int itself I feel

anmolbabu: Also I think it might be better to call it quorum instead of flag so that if any more such flags are added the name is online with purpose..

tendrl/commons/utils/etcd_utils.py
```diff
@@ -0,0 +1,68 @@
...
+ 
+def read(key):
+    try:
+        return NS._int.client.read(key)
+    except (etcd.EtcdConnectionFailed, etcd.EtcdException) as ex:
+        if type(ex) != etcd.EtcdKeyNotFound:
+            NS._int.reconnect()
+            return NS._int.client.read(key)
+        else:
+            return None
```
anmolbabu: Better to raise exception to be able to differentiate between actual Null value and absence of key

tendrl/commons/utils/etcd_utils.py
```diff
@@ -0,0 +1,68 @@
...
+
+   return param:
+       dict - > if read is successful
+       None - > if key is not found
+'''
+ 
+def read(key):
+    try:
+        return NS._int.client.read(key)
+    except (etcd.EtcdConnectionFailed, etcd.EtcdException) as ex:
```
anmolbabu: I have seen cases where when reads and/or writes by multiple tendrl processes failed (almost) parallelly, each of them try a reconnect. And what happens in reconnect is the client is first reset to null and then set to a new etcd client.. So what I feel is the best and/or easy approach is,  to rather gracefully accept AttributeError(and raise it off course so that the users of the utility can custom handle it) As this is expected to happen consecutive ly only a fixed number of times (equal to number of tendrl processes parallelly making those reads and/or writes) and this I feel will not land us into any sort of infinite maze :-) or not even will it cause any race condition...

tendrl/commons/utils/etcd_utils.py
rishubhjain: Done :)

tendrl/commons/utils/etcd_utils.py
anmolbabu: So, as we discussed this won't be required as the handler can still catch such exceptions.... even with the current approach...

rishubhjain:
@r0h4n Added , https://travis-ci.org/Tendrl/commons/jobs/246137004#L277

PR #627, opened 2017-06-22
ExpandCluster flow should be of type Update and not Create

nnDarshan:


nnDarshan:
@r0h4n @nthomas-redhat @anivargi @shtripat Please review

PR #631, opened 2017-06-27
Added cluster network details entity

shtripat:
tendrl-bug-id: Tendrl/commons#630

### DIFF
--- a/tendrl/commons/flows/create_cluster/__init__.py
+++ b/tendrl/commons/flows/create_cluster/__init__.py
@@ -202,6 +202,13 @@ class CreateCluster(flows.BaseFlow):
                 self.parameters
             )
 
+            # Persist the cluster's extra details
+            NS.tendrl.objects.Cluster(
+                integration_id=self.parameters['TendrlContext.integration_id'],
+                public_network=self.parameters.get('Cluster.public_network'),
+                cluster_network=self.parameters.get('Cluster.cluster_network')
+            ).save()
+
             Job(job_id=_job_id,
                 status="new",
                 payload=payload).save()
--- a/tendrl/commons/objects/definition/master.yaml
+++ b/tendrl/commons/objects/definition/master.yaml
@@ -83,16 +83,20 @@ namespace.tendrl:
 
   objects:
     Cluster:
-      internal: True
       enabled: True
       help: "Represents a cluster"
       list: /clusters
+      value: clusters/$TendrlContext.integration_id
       attrs:
+        integration_id:
+          help: integration id
+          type: String
         public_network:
           help: Public Network cidr of the cluster
           type: String
         cluster_network:
           help: Data Nework cidr of the cluster
+          type: String
         node_configuration:
           help: Node configuration for the cluster nodes
           type: Dict[]
@@ -101,7 +105,7 @@ namespace.tendrl:
           type: Dict
         node_identifier:
           help: How nodes should be identified while create cluster
-          type: String
+          type: list
     DetectedCluster:
       enabled: True
       help: "DetectedCluster"
--- a/tendrl/commons/objects/cluster/__init__.py
+++ b/tendrl/commons/objects/cluster/__init__.py
@@ -1,0 +1,22 @@
+from tendrl.commons import objects
+
+
+class Cluster(objects.BaseObject):
+    def __init__(self, integration_id=None, public_network=None,
+                 cluster_network=None, node_configuration=None,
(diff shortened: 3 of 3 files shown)


shtripat:
Not verified yet

tendrl/commons/objects/cluster_network_details/__init__.py
```diff
@@ -0,0 +1,17 @@
+from tendrl.commons import objects
+
+
+class ClusterNetworkDetails(objects.BaseObject):
```
r0h4n: There is already a "Cluster" object, implement it here https://github.com/Tendrl/commons/blob/develop/tendrl/commons/objects/definition/master.yaml#L85

tendrl/commons/objects/definition/master.yaml
```diff
@@ -669,4 +669,18 @@ namespace.tendrl:
       help: "Platform of the Node"
       value: nodes/$NodeContext.node_id/Platform
       list: nodes/$NodeContext.node_id/Platform
+    ClusterNetworkDetails:
```
r0h4n: Use https://github.com/Tendrl/commons/blob/develop/tendrl/commons/objects/definition/master.yaml#L85

tendrl/commons/objects/cluster_network_details/__init__.py
shtripat: Done

tendrl/commons/objects/definition/master.yaml
shtripat: Done

shtripat:
Verified the changes with create cluster flow of gluster cluster. Works as expected and the details are avvailable at `clusters/{int-id}/public_network` and `clusters/{int-id}/cluster_network`. In case of glcutsre cluster only `clusters/{int-id}/cluster_network` would be populatd as passed while create cluster flow.

@gnehapk ^^^

tendrl/commons/flows/create_cluster/__init__.py
```diff
@@ -232,3 +232,10 @@ def run(self):
         finally:
             # releasing nodes if any exception came
             create_cluster_utils.release_node_lock(self.parameters)
+
+            # Persist the cluster's extra details
+            NS.tendrl.objects.Cluster(
```
r0h4n: This finally block can be reached even if there is an exception, do you really want to save network details in case of exception?

tendrl/commons/objects/definition/master.yaml
```diff
@@ -83,11 +83,14 @@ namespace.tendrl:
...
     Cluster:
-      internal: True
       enabled: True
       help: "Represents a cluster"
       list: /clusters
+      value: clusters/$TendrlContext.integration_id
       attrs:
+        integration_id:
+          help: integration id
+          type: String
```
r0h4n: node_identifier type should be "list"

tendrl/commons/flows/create_cluster/__init__.py
```diff
@@ -232,3 +232,10 @@ def run(self):
...
+                integration_id=self.parameters['TendrlContext.integration_id'],
+                public_network=self.parameters.get('Cluster.public_network', ""),
+                cluster_network=self.parameters.get('Cluster.cluster_network', "")
```
r0h4n: Cluster.public_network and Cluster.cluster_network should not be set as empty

tendrl/commons/flows/create_cluster/__init__.py
shtripat: May be this could be done just after releasing the node locks, as that's the place where create is done at least.

tendrl/commons/flows/create_cluster/__init__.py
shtripat: In case of gluster cluster creation only cluster_network value is available in parameters. May be we can add mandatory availability of cluster_network at least. Also by this time as cluster creation is done so at least cluster_network should be available.

tendrl/commons/objects/definition/master.yaml
shtripat: Why `list` and how we would call load() in that case? Currently load is something like

```
NS.tendrl.objects.Cluster(integration_id=NS.tendrl_context.integration_id).load()
```

How this would work in case we have identifier as `list`?

shtripat: Confused `integration_id` with `node_identifier`. Changed the type of `node_identifier` as list.

Issue #603, opened 2017-06-16
Use tags instead of node_ids for targetting job to specific node


r0h4n:
Currently we use "node_ids: [node_id]"  in Job payload to target job to a specific node

The node-agent now tags itself with "tendrl/node_$node_id" 

eg: "tendrl/node_34asd22asdadsasasd" 

We need to stop using "node_ids" in Job payload and replace it with eg:  "tags: ["tendrl/node_34asd22asdadsasasd"] while creating such Jobs anywhere across Tendrl repos

r0h4n:
@anmolbabu @nnDarshan @shtripat please get start on this change for your repos

r0h4n:
@anmolbabu @nnDarshan @shtripat Any updates?

shtripat:
@r0h4n I remember while create and import cluster flows we are using `node_id` for targetting the jobs to specific nodes. I will work on a PR to fix that and update here.

r0h4n:
@shtripat Is this done?

shtripat:
Yes the patches were merged.

PR #636, opened 2017-06-28
Modified to use tags instead of node_ids for jobs

shtripat:
tendrl-bug-id: Tendrl/commons#603


shtripat:
Needs to be verified

shtripat:
Verified the changes with create/import and expand flows. Works fine.

PR #652, opened 2017-07-20
Update __init__.py

shtripat:
tendrl-bug-id: Tendrl/commons#654


shtripat:
@r0h4n this is a regression. Kindly merge.

PR #670, opened 2017-08-04
Modified to get minor version details for the case `3.8dev.0`

shtripat:
The existing code doesnt parse the minor version properly for
above case. Updated to handle the case properly.


r0h4n:
Verified?

shtripat:
Not verified yet. Verified the `re.findall()` function with sample data `3.8dev.0` though :)

PR #674, opened 2017-08-04
Modified to get minor version details for the case `3.8dev.0`

shtripat:
The existing code doesnt parse the minor version properly for
above case. Updated to handle the case properly.


shtripat:
Verified the changes with glusterfs nightly build (version 4.0dev) and 3.11 (version 3.11) and works fine.

PR #675, opened 2017-08-04
Disabled gdeploy install/configuration

shtripat:
tendrl-bug-id: Tendrl/commons#679


shtripat:
Verified the import cluster flow with changes and it works fine.

nnDarshan:
Would it be better if we have a flag in the yaml definition file to control the installation instead of removing the code... as this would be definitely needed in the future releases of tendrl

shtripat:
@nnDarshan thats a good idea. Anyway changes are not huge and they can be brought back later when required. @Tendrl/tendrl-core comments??

r0h4n:
@shtripat Please attach an issue for this. 

Also, we will provide gdeploy auto install when it is required, right now It is not required.

Issue #655, opened 2017-07-24
The load_all() function returns None


shtripat:
While trying to load all the volumes as below

```
volumes = NS.gluster.objects.Volume().load_all()
```

it returns `None`. It should return all the volumes for the current cluster correctly.

r0h4n:
The fix linked to this bug looks like a hack, can we discuss a better approach here?

r0h4n:
@shtripat @rishubhjain updates?

rishubhjain:
I am debugging the issue and trying to find a solution, but for now I don't have a better solution. Will try to find a better approach as soon as possible

r0h4n:
@rishubhjain whats the update on this? I see a PR from @shtripat

rishubhjain:
@r0h4n I don't have a better solution for now. I am working on Monitoring-Integration. Will it be ok if I could work on this after completing monitoring-integration task?

PR #685, opened 2017-08-11
Fixed a regression issue with load_all() function

shtripat:
tendrl-bug-id: Tendrl/commons#655


tendrl/commons/objects/__init__.py
```diff
@@ -201,7 +202,9 @@ def load(self):
                 pass
 
         _copy = self._copy_vars()
-
+        # Check if self.value already set, use it
+        if self.value.find('{') < 0:
```
r0h4n: self.render is already called on L167. It safe to assume self.value is rendered correctly, no need to check

shtripat: `_copy_vars()` sets the old `value` field again in `_copy`. So I check if self.value doesn't contain `{` it means its properly rendered and so I assign it back to `_copy` variable's `value` field.

r0h4n: Well you can include the rendered self.value in _copy_vars method, but dont do this self.value.find

shtripat: Including `value` field in `_copy_vars()` causes an error `TypeError: __init__() got an unexpected keyword argument 'value'` while normal `load()` calls for entities, so took this path :) Just changed `_copy_vars` and saw this error again.

shtripat: May be we can copy `self.value` un-conditionally to `_copy`

Issue #690, opened 2017-08-17
The "inputs" field should be optional for flows


shtripat:
In some scenarios, the flows might not need `inputs` and so this field should be optional.

PR #691, opened 2017-08-17
Made inputs field as optional for flows

shtripat:
tendrl-bug-id: Tendrl/commons#690


shtripat:

PR #692, opened 2017-08-17
Mark the node which picks the import cluster job as provisioner

shtripat:
The first node which picks the import cluster job would be marked
with tag `provisioner/gluster`. Later while syncing the cluster
details this tag would be used for creating aggregated sync job.

tendrl-bug-id: Tendrl/gluster-integration#359

### DIFF
--- a/tendrl/commons/objects/cluster/atoms/import_cluster/__init__.py
+++ b/tendrl/commons/objects/cluster/atoms/import_cluster/__init__.py
@@ -152,6 +152,8 @@ class ImportCluster(objects.BaseAtom):
                             )
                         )
                     else:
+                        # TODO(shtripat) ceph-installer is auto detected and provisioner/$integration_id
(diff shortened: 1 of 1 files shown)


shtripat:

tendrl/commons/objects/cluster/atoms/import_cluster/__init__.py
```diff
@@ -151,6 +151,9 @@ def run(self):
                                 }
                             )
                         )
+                    else:
+                        NS.node_context.tags += ['provisioner/%s' % integration_id]
```
r0h4n: Should not set this tag for ceph, since ceph-installer is pre-installed by the user.on any cluster node and tendrl detects the ceph-installer service and tags that node "provisioner/$integration_id"

Please add a todo here to handle this case here

Issue #683, opened 2017-08-09
Add "errors" to NS.tendrl.objects.Cluster


r0h4n:
During SDS detection if there are errors, append them as string to Cluster.errors.

Also set sds_version and name (i.e $sds_name_short_integration_id)
Tendrl definitions need to be updated for the same

Issue #682, opened 2017-08-09
Add "enable_volume_profiling" attribute to NS.tendrl.objects.Cluster


r0h4n:
This flag enables gluster volume profiling for existing volumes and any volume created as long as the flag is enabled.

This flag is provided as input to tendrl-api ImportCluster call. Tendrl definitions for this flow will be updated

shtripat:
@r0h4n this could be closed

PR #695, opened 2017-08-22
Import cluster improvements

r0h4n:
tendrl-bug-id: Tendrl/commons#682
tendrl-bug-id: Tendrl/commons#683
tendrl-bug-id: Tendrl/specifications#211
tendrl-bug-id: Tendrl/specifications#213


r0h4n:
@Tendrl/tendrl-core @shtripat @nnDarshan 

Note: setting up of password less ssh is removed since Tendrl doesnt install gdeploy (it is now a pre-requisite on users to install gdeploy )

PR #711, opened 2017-08-31
Added utility to update dashboard with required details

shtripat:
tendrl-bug-id: Tendrl/gluster-integration#385

### DIFF
--- a/tendrl/commons/utils/monitoring_utils.py
+++ b/tendrl/commons/utils/monitoring_utils.py
@@ -1,0 +1,27 @@
+import uuid
+
+from tendrl.commons.objects.job import Job
+
+
+def update_dashboard(res_name, res_type, integration_id, action):
+    _job_id = str(uuid.uuid4())
+    _params = {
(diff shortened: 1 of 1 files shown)


shtripat:
This is required by https://github.com/Tendrl/gluster-integration/pull/386.
Its done as part of commons because, this functionality might be required by other integration module like `ceph-integration` later.

tendrl/commons/utils/monitoring_utils.py
```diff
@@ -0,0 +1,25 @@
...
+        "run": "monitoring.flows.UpdateDashboard",
+        "status": "new",
+        "parameters": _params,
+        "type": "monitoring"
+    }
+    Job(
+        job_id=_job_id,
+        status="new",
+        payload=_job_payload
+    ).save()
```
r0h4n: return _job_id

shtripat: Done

PR #715, opened 2017-09-03
Plugin config changes as per new plugin architecture

anmolbabu:

### DIFF
--- a/tendrl/commons/objects/cluster/atoms/configure_monitoring/__init__.py
+++ b/tendrl/commons/objects/cluster/atoms/configure_monitoring/__init__.py
@@ -24,13 +24,7 @@ NODE_PLUGINS = {
 
 
 GLUSTER_CLUSTER_PLUGINS = {
-    'node_plugins': {
-        'tendrl_glusterfs_health_counters',
-        'tendrl_glusterfs_brick_utilization'
-    },
-    'cluster_plugins': {
-        'tendrl_glusterfs_profile_info'
-    }
+    'tendrl_gluster'
 }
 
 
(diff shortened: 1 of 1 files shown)


anmolbabu:
Needs to be verified

tendrl/commons/objects/cluster/atoms/configure_monitoring/__init__.py
```diff
@@ -117,17 +111,12 @@ def run(self):
...
             ) in NS.node_context.tags:
-                plugins.update(
-                    GLUSTER_CLUSTER_PLUGINS.get(
-                        'cluster_plugins',
-                        []
-                    )
-                )
-            for gluster_plugin in plugins:
+                plugin_params['is_provisioner_node'] = True
+            for gluster_plugin in GLUSTER_CLUSTER_PLUGINS:
```
shtripat: So is it like only provisioner node would be configured with collectd plugins? Because in parameters I see current node FQDN only...

anmolbabu: No no.
Every node will have health counters and brick utilization plugins configured but only 
The way it works is as follows:
1. All plugins are installed on all gluster cluster nodes as part of post atom in import cluster flow.
2. The config on every node will contain a field called 'is_provisioner_node' with possible values as True or False to indicate if current node is provisioner node.
3. Every plugin on the other hand will also have a field(boolean) 'provisioner_only_plugin' to indicate if the specific plugin is a provisioner node only plugin.
4. Using 2 and 3 appropriate plugins are only intiated and run by our base framework within collectd.

Note, the generic parameters are formed as in https://github.com/Tendrl/commons/pull/715/files#diff-059917ac9f73bcf3a89b77a6b4806857L104 and the provisioner flag for config is set as in https://github.com/Tendrl/commons/pull/715/files#diff-059917ac9f73bcf3a89b77a6b4806857R114 and 
https://github.com/Tendrl/commons/pull/715/files#diff-059917ac9f73bcf3a89b77a6b4806857R118

anmolbabu:
@nnDarshan @nthomas-redhat @r0h4n Please review

tendrl/commons/objects/cluster/atoms/configure_monitoring/__init__.py
shtripat: Ack

PR #716, opened 2017-09-05
Modified alert objects to read by load function

GowthamShanmugam:
tendrl-spec-issue: Tendrl/specifications#169

### DIFF
--- a/tendrl/commons/objects/cluster_alert/__init__.py
+++ b/tendrl/commons/objects/cluster_alert/__init__.py
@@ -15,8 +15,8 @@ class ClusterAlert(Alert, objects.BaseObject):
     def render(self):
         try:
             self.tags = json.loads(self.tags)
-        except TypeError:
-            # Already in json format
+        except(TypeError, ValueError):
+            # Already in json or None
             pass
         self.value = self.value.format(
             self.tags['integration_id'],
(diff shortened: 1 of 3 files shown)


GowthamShanmugam:
dumping tags in the alert object as a string to use load() and load_all() method instead of direct etcd read

GowthamShanmugam:
@r0h4n please review

tendrl/commons/objects/cluster_alert/__init__.py
```diff
@@ -12,7 +13,13 @@ def __init__(self, *args, **kwargs):
         self.value = 'alerting/clusters/{0}/{1}'
 
     def render(self):
-        self.value = self.value.format(self.tags['integration_id'],
-                                       self.alert_id
-                                       )
+        try:
+            self.tags = json.loads(self.tags)
+        except TypeError:
```
r0h4n: json.loads raises ValueError and not a TypeError, why have you used TypeError?

GowthamShanmugam: @r0h4n For none value and if value already json then it will raise TypeError also , it will add both

GowthamShanmugam:
@r0h4n changes done

Issue #717, opened 2017-09-05
Load function not works for cluster tendrl context object 


GowthamShanmugam:
As per load method implementation in the _copy_vars function, object attributes are passed to the constructor as kwargs. The problem here is parameter name for integration_id in constructor does not match with attribute name in an object so kwargs can't assign a value for that. Also, integration_id is used in render method to form values, Because of this problem integration_id always comes none.

The parameter names should always same as object attributes, then only load will work fine.

GowthamShanmugam:
@r0h4n @shtripat

shtripat:
If that's the case shouldn't we correct the attribute name in class?

GowthamShanmugam:
@shtripat only param is the problem, param should have same which is present in attribute name in class

GowthamShanmugam:
This issue is solved i am closing this issue

PR #718, opened 2017-09-05
Modified cluster tendrl context object parameter

GowthamShanmugam:
tendrl-bug-id: Tendrl/commons#717


GowthamShanmugam:
@shtripat @r0h4n please revie

GowthamShanmugam:
In all other places it is called as integration_id but in constructor param the name is int_id, i have checked all other places only fix required here.

PR #719, opened 2017-09-05
Fix for alert object

GowthamShanmugam:
tendrl-bug-id: Tendrl/commons#717

Small fixed missed

Signed-off-by: root <root@dhcp43-56.lab.eng.blr.redhat.com>


GowthamShanmugam:
@r0h4n please merge this

PR #720, opened 2017-09-06
Added alerting mechanism for job status

shtripat:
tendrl-bug-id: Tendrl/node-agent#602

### DIFF
--- a/tendrl/commons/jobs/__init__.py
+++ b/tendrl/commons/jobs/__init__.py
@@ -14,6 +14,7 @@ from tendrl.commons.message import ExceptionMessage
 from tendrl.commons.message import Message
 from tendrl.commons.objects import AtomExecutionFailedError
 from tendrl.commons.objects.job import Job
+from tendrl.commons.utils import alert_utils
 from tendrl.commons.utils import time_utils
 
 
@@ -104,6 +105,19 @@ def process_job(job):
                     _msg = str("Timed-out (>10min as 'new')")
                     job.errors = _msg
                     job.save()
+                    if job.payload.get('parent') is None:
+                        alert_utils.alert_job_status(
+                            "failed",
+                            "Job timed out (job_id: %s)" % jid,
+                            integration_id=NS.tendrl_context.integration_id or
+                            job.payload['parameters'].get(
+                                'TendrlContext.integration_id'
+                            ),
+                            cluster_name=NS.tendrl_context.cluster_name or
+                            job.payload['parameters'].get(
+                                'TendrlContext.cluster_name'
+                            )
+                        )
                     return
         else:
             _now_plus_10 = time_utils.now() + datetime.timedelta(minutes=10)
@@ -219,6 +233,19 @@ def process_job(job):
                              }
                 )
             )
+            if job.payload.get('parent') is None:
+                alert_utils.alert_job_status(
+                    "finished",
+                    "Job finished successfully (job_id: %s)" % job.job_id,
+                    integration_id=NS.tendrl_context.integration_id or
+                    job.payload['parameters'].get(
+                        'TendrlContext.integration_id'
+                    ),
+                    cluster_name=NS.tendrl_context.cluster_name or
+                    job.payload['parameters'].get(
+                        'TendrlContext.cluster_name'
+                    )
+                )
         except (FlowExecutionFailedError,
                 AtomExecutionFailedError,
                 Exception) as e:
@@ -265,6 +292,19 @@ def process_job(job):
             else:
                 job = job.load()
                 job.errors = _trace
+                if job.payload.get('parent') is None:
+                    alert_utils.alert_job_status(
+                        "failed",
+                        "Job failed (job_id: %s)" % job.job_id,
+                        integration_id=NS.tendrl_context.integration_id or
+                        job.payload['parameters'].get(
+                            'TendrlContext.integration_id'
+                        ),
+                        cluster_name=NS.tendrl_context.cluster_name or
+                        job.payload['parameters'].get(
+                            'TendrlContext.cluster_name'
+                        )
+                    )
                 job.save()
 
 
--- a/tendrl/commons/utils/alert_utils.py
+++ b/tendrl/commons/utils/alert_utils.py
@@ -1,0 +1,38 @@
+import json
+import os
+
+from tendrl.commons.utils import log_utils as logger
+from tendrl.commons.utils.time_utils import now as tendrl_now
+
+
+def alert_job_status(curr_value, msg, integration_id=None, cluster_name=None):
+    alert = {}
+    alert['source'] = NS.publisher_id
+    alert['classification'] = 'cluster'
+    alert['pid'] = os.getpid()
+    alert['time_stamp'] = tendrl_now().isoformat()
+    alert['alert_type'] = 'STATUS'
+    severity = "INFO"
+    if curr_value.lower() == "failed":
+        severity = "WARNING"
+    alert['severity'] = severity
+    alert['resource'] = 'job_status'
+    alert['current_value'] = curr_value
+    alert['delivered'] = True
+    alert['tags'] = dict(
+        message=msg,
+        integration_id=integration_id or
+        NS.tendrl_context.integration_id,
+        cluster_name=cluster_name or
+        NS.tendrl_context.cluster_name,
+        sds_name=NS.tendrl_context.sds_name,
+        fqdn=NS.node_context.fqdn
+    )
+    alert['node_id'] = NS.node_context.node_id
+    if not NS.node_context.node_id:
+        return
+    logger.log(
+        "notice",
+        "alerting",
+        {'message': json.dumps(alert)}
+    )
--- a/tendrl/commons/utils/alert_utils.py
+++ b/tendrl/commons/utils/alert_utils.py
@@ -18,7 +18,7 @@ def alert_job_status(curr_value, msg, integration_id=None, cluster_name=None):
     alert['severity'] = severity
     alert['resource'] = 'job_status'
     alert['current_value'] = curr_value
-    alert['delivered'] = True
+    alert['notify_only_ui'] = True
     alert['tags'] = dict(
         message=msg,
         integration_id=integration_id or
--- a/tendrl/commons/utils/monitoring_utils.py
+++ b/tendrl/commons/utils/monitoring_utils.py
@@ -7,6 +7,7 @@ def update_dashboard(res_name, res_type, integration_id, action):
     _job_id = str(uuid.uuid4())
     _params = {
         "TendrlContext.integration_id": NS.tendrl_context.integration_id,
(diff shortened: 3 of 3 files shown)


shtripat:
This is not yet verified.

tendrl/commons/utils/alert_util.py
```diff
@@ -0,0 +1,38 @@
...
+        message=msg,
+        cluster_id=NS.tendrl_context.integration_id,
+        cluster_name=NS.tendrl_context.cluster_name,
+        sds_name=NS.tendrl_context.sds_name,
+        fqdn=socket.getfqdn()
+    )
+    alert['node_id'] = NS.node_context.node_id
+    if not NS.node_context.node_id:
+        return
+    Event(
```
r0h4n: use logger utils

tendrl/commons/jobs/__init__.py
```diff
@@ -104,6 +105,10 @@ def process_job(job):
                     _msg = str("Timed-out (>10min as 'new')")
                     job.errors = _msg
                     job.save()
+                    alert_utils.alert_job_status(
+                        "failed",
+                        "Job %s timed out" % jid
```
r0h4n: msgs which contain any ID's or uuids should be framed like below

"Job timed out (job_id: %jid)" 

same for other msgs below

tendrl/commons/utils/alert_util.py
shtripat: Ok. Will check and use that

tendrl/commons/jobs/__init__.py
shtripat: Done

shtripat:
@r0h4n there is a case where flow from `monitoring_integration` is invoked for updating dashboards. In case of `monitoring_integration` we don't have `NS.tendrl_context.integration_id` populated as this is server node. This field is required for raising the cluster level alerts.

As a work around we can take the value from `job.payload['parameters']['TendrlContext.integration_id']` and pass to `alert_utils.alert_job_status()` but still `cluster_name` field would be missing.

What you suggest?

tendrl/commons/utils/alert_util.py
shtripat: I am not sure log_utils could be used for `notice` type. Still would verify once with @GowthamShanmugam and @rishubhjain on this.

tendrl/commons/jobs/__init__.py
r0h4n: "Job timed out (job_id: %jid)"

This is how the msg should look like, I dont see it changed, please change it

tendrl/commons/utils/alert_utils.py
```diff
@@ -0,0 +1,40 @@
...
+        severity = "WARNING"
+    alert['severity'] = severity
+    alert['resource'] = 'cluster'
+    alert['current_value'] = curr_value
+    alert['tags'] = dict(
+        message=msg,
+        integration_id=NS.tendrl_context.integration_id,
+        cluster_name=NS.tendrl_context.cluster_name,
+        sds_name=NS.tendrl_context.sds_name,
+        fqdn=socket.getfqdn()
```
r0h4n: use NS.node_context.fqdn

tendrl/commons/utils/alert_utils.py
```diff
@@ -0,0 +1,40 @@
+import json
+import os
+import socket
```
r0h4n: remove socket

tendrl/commons/utils/alert_utils.py
```diff
@@ -0,0 +1,40 @@
...
+
+from tendrl.commons.event import Event
+from tendrl.commons.message import Message
+from tendrl.commons.utils.time_utils import now as tendrl_now
+
+
+def alert_job_status(curr_value, msg):
+    import pdb; pdb.set_trace()
```
r0h4n: remove the pdb

tendrl/commons/utils/alert_utils.py
```diff
@@ -0,0 +1,40 @@
...
+    alert['source'] = NS.publisher_id
+    alert['classification'] = 'cluster'
+    alert['pid'] = os.getpid()
+    alert['time_stamp'] = tendrl_now().isoformat()
+    alert['alert_type'] = 'STATUS'
+    severity = "INFO"
+    if curr_value.lower() == "failed":
+        severity = "WARNING"
+    alert['severity'] = severity
+    alert['resource'] = 'cluster'
```
GowthamShanmugam: In resource, you have to give like job status or like volume status based on the alert type

GowthamShanmugam: in future, if we allow the user choose what are all the alert types they want to raise the alert then we can restrict are allow the tendrl by using this field only.  These resource fields act as alert types in etcd.

under utilization alert types : brick utilization, cpu utilization ..etc these will come. so resource are the sub alert types under alert types

tendrl/commons/utils/alert_utils.py
```diff
@@ -0,0 +1,40 @@
...
+        integration_id=integration_id or
+        NS.tendrl_context.integration_id,
+        cluster_name=cluster_name or
+        NS.tendrl_context.cluster_name,
+        sds_name=NS.tendrl_context.sds_name,
+        fqdn=NS.node_context.fqdn
+    )
+    alert['node_id'] = NS.node_context.node_id
+    if not NS.node_context.node_id:
+        return
```
GowthamShanmugam: @shtripat may i know why you are setting node_id for cluster related alert?

tendrl/commons/jobs/__init__.py
```diff
@@ -219,6 +230,16 @@ def process_job(job):
                              }
                 )
             )
+            alert_utils.alert_job_status(
+                "finished",
+                "Job finished successfully (job_id: %s)" % job.job_id,
+                integration_id=job.payload['parameters'].get(
+                    'TendrlContext.integration_id'
```
r0h4n: Raise alert only when  `job.payload['parent'] is None` to ensure that this is not a child job.

tendrl/commons/jobs/__init__.py
```diff
@@ -219,6 +230,16 @@ def process_job(job):
                              }
                 )
             )
+            alert_utils.alert_job_status(
+                "finished",
+                "Job finished successfully (job_id: %s)" % job.job_id,
+                integration_id=job.payload['parameters'].get(
```
r0h4n: Fetch integration id and cluster name from NS.tendrl_context, if NS.tendrl_context is not available only then look in the job.payload['parameters'] for integration_id and cluster_name

tendrl/commons/utils/alert_utils.py
shtripat: As commented earlier by @r0h4n node_id should be made mandatory

tendrl/commons/jobs/__init__.py
shtripat: Done

tendrl/commons/jobs/__init__.py
shtripat: Done

tendrl/commons/utils/alert_utils.py
shtripat: I think I tried with `job_status` but faced some issue. Anyway will try once and change if required.

tendrl/commons/utils/alert_utils.py
```diff
@@ -0,0 +1,40 @@
...
+    Event(
+        Message(
+            "notice",
```
nthomas-redhat: Why you want to send out alert emails for job status updates? What is expected here is a notification, we should not send out alert emails for this.

shtripat: @nthomas-redhat its in accordance with https://github.com/Tendrl/node-agent/issues/602. We need notifications in case of top level job success/failures

nthomas-redhat: @shtripat , we need notifications, not alerts(emails or SNMP). If you set the priority to notice, alerts will be going out

r0h4n: @nthomas-redhat how do you send a notification instead of alert?

shtripat: @GowthamShanmugam ^^^

nthomas-redhat:
> @nthomas-redhat how do you send a notification instead of alert?

@r0h4n , @shtripat , you need to set a priority other than `notice`; i.e. , critical, error, warning and info
@GowthamShanmugam , please confirm this.

tendrl/commons/utils/alert_utils.py
```diff
@@ -0,0 +1,37 @@
...
+        cluster_name=cluster_name or
+        NS.tendrl_context.cluster_name,
+        sds_name=NS.tendrl_context.sds_name,
+        fqdn=NS.node_context.fqdn
+    )
+    alert['node_id'] = NS.node_context.node_id
+    if not NS.node_context.node_id:
+        return
+    logger.log(
+        "notice",
```
r0h4n: This shouldnt be "notice" , it should be either "info" if job succeeded or "error" if job failed/timedout

shtripat: @r0h4n I remember if its not `notice` the data goes to `messages/events` only and not to `alerting/clusters/{int-id}`.

shtripat: After discussion with @GowthamShanmugam and looking at api implementation at https://github.com/Tendrl/api/blob/master/app/models/notification.rb, we find that notifications drawer menu in UI gets data from `messages/events` but only messages with `log_priorty = notice` are listed here.

So the changes in this PR remain required. As per @GowthamShanmugam to supress, email alerts for these, we can set `delivered = True` for alert message payload. This has a drawback (but not braking any functionality) that the job success/failure would be listed under Alerts as well. 

@GowthamShanmugam needs a fix in node-agent's alert module to expect another parameter in alert payload to not put the `notice` type message in `/alerting` and it would meant for only notifications. Till this change comes, we would go ahead with parameter `delivered = True` to achieve this.

Will be updating the PR with this soon. Hope this ends the long going discussion around this :) no offences ;)

GowthamShanmugam:
@shtripat @nthomas-redhat @r0h4n as per current implementation, if message has Notice priority then I am considering the message as alert message, and only message which has Notice priority is stored in /alerting/alerts as well as /messages/events and notified by email. Other messages are stored in only in  /messages/events.

So now as per current discussion alert also have Notice priority and also notification also have Notice priority
so only alert message which has Notice priority should go to alerting/alerts and /messages/events as well as notified by email , and notification message which has notice priority should go only /messages/event. 

for this, i need to add some extra attr in the payload to identify message which has notice priority is alert or notification. So for only alert, i am going to give some extra attr in the message payload. if those attr present then node_agent can understand it is alert message. if it is not present then node_agent can understand it is notification message.

Issue #723, opened 2017-09-09
Breakage in cluster import


nthomas-redhat:
Failure in Job 4eb4584b-8abb-40e6-a64e-6e2bbd2b2e6d Flow tendrl.flows.ImportCluster with error: Traceback (most recent call last): File "/usr/lib/python2.7/site-packages/tendrl/commons/jobs/__init__.py", line 198, in process_job the_flow.run() File "/usr/lib/python2.7/site-packages/tendrl/commons/flows/import_cluster/__init__.py", line 35, in run raise FlowExecutionFailedError("Cluster already being imported by another Job, please wait till the job finishes (job_id: %s) (integration_id: %s) " % (_cluster.import_job_id, _cluster.integration_id)) FlowExecutionFailedError: Cluster already being imported by another Job, please wait till the job finishes (job_id: ) (integration_id: 7673bb91-8f11-43b5-b8d0-3fec3ea196fb)

r0h4n:
Fixed: https://github.com/Tendrl/commons/commit/b57f3bb43010e93c2c2b3643ee464e48c281c57a

Fixed: https://github.com/Tendrl/commons/commit/24b7e0c3634e3ecf1722b66d2db7b6556a15275a

PR #724, opened 2017-09-09
Fix import issue

nthomas-redhat:
tendrl-bug-id:Tendrl/commons#723


tendrl/commons/flows/import_cluster/__init__.py
```diff
@@ -21,15 +21,28 @@ def run(self):
...
                 pass
             
             try:
 
-                NS._int.wclient.write(_cluster_import_status, "in_progress", prevExist=False)
-                NS._int.wclient.write(_cluster_import_job_id, self.job_id, prevExist=False)
+                NS._int.wclient.write(
+                    _cluster_import_status,
+                    "in_progress",
+                    prevExist=True)
```
r0h4n: prevExist=True means  this mark import_status as "in_progress" even if it is "done", "in_progress" or "failed"

Why would you even want this? This code is about locking a cluster import, your change doesnt do anything related to that

r0h4n:
Fixed : https://github.com/Tendrl/commons/issues/723#issuecomment-328323346

PR #725, opened 2017-09-10
Add etcd params to config + disk stats plugin

anmolbabu:


anmolbabu:
Verified

anmolbabu:

r0h4n:
File an issue, are the collectd plugins now gonna talk to etcd?

anmolbabu:
@r0h4n Yes the collectd plugin to link disk and brick talks to etcd because the blivet behaviour from within collectd plugin was not very consistent from within collectd plugins as briefed in https://github.com/Tendrl/node-agent/issues/611 point 2.So this is going to be the only plugin talking to etcd. This lugin has been tested to work with gluster-integration updated yesterday from master repo.

PR #727, opened 2017-09-11
Update alert_utils.py

shtripat:
Correctly passing the flag `notify_only_ui` under message payload.


shtripat:
@r0h4n this is a regression. Kindly merge this.

PR #729, opened 2017-09-12
Added new objects for sds native alert

GowthamShanmugam:
This kind of alerts does not have clearing alert so
it is never comes under alert condtion. so we are
storing these aler in /alerting/notify.

tendrl-bug-id : Tendrl/commons#730

### DIFF
--- a/tendrl/commons/objects/definition/master.yaml
+++ b/tendrl/commons/objects/definition/master.yaml
@@ -759,7 +759,7 @@ namespace.tendrl:
       value: alerting/clusters/$Alert.tags['integration_id']/$Alert.alert_id
       list: alerting/clusters/$Alert.tags['integration_id']
       help: "Cluster alerts"
-    UtilizationOnlyAlert:
+    NotificationOnlyAlert:
       attrs:
         alert_id:
           help: 'The unique identifier of alert'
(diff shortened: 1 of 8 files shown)


shtripat:
Where is the logic for persisting these alerts? Also I think rather we should call this as `UtilizationAlert` which is meant for saving alerts for which there are not going to be clearing events.

GowthamShanmugam:
@shtripat modified

GowthamShanmugam:
@shtripat logic is in https://github.com/Tendrl/node-agent/pull/618

tendrl/commons/objects/definition/master.yaml
```diff
@@ -759,6 +759,63 @@ namespace.tendrl:
       value: alerting/clusters/$Alert.tags['integration_id']/$Alert.alert_id
       list: alerting/clusters/$Alert.tags['integration_id']
       help: "Cluster alerts"
+    UtilizationOnlyAlert:
```
r0h4n: What is the connection with "Utilization" in the name? @shtripat @GowthamShanmugam

GowthamShanmugam: @shtripat  i am also felt this name will confuse.. need to change this object name

shtripat: Ah my bad I meant `NotificationOnlyAlert` and not `UtilizationOnlyAlert` :(
This alert class represents alerts for which no clearing events would come later.

Issue #721, opened 2017-09-08
Enable ETCD authorization and authentication


r0h4n:
Tendrl services should use ETCD certificate based auth for any communications with ETCD

https://github.com/coreos/etcd/blob/master/Documentation/authentication.md

Related issue: https://github.com/Tendrl/api/issues/256

r0h4n:
- Etcd User/Password based auth is enabled
- Etcd TLS cert based auth in progress, will be done soon

r0h4n:
Seems like ETCD basicauth (user/password) adds a performance penalty in API responses, we will work to implement TLS cert based auth asap

mbukatov:
@sidhax this seems like security related issue, here we need to figure out how to use etcd cert auth instead of password based auth

r0h4n:
This is fixed

Please check https://github.com/Tendrl/documentation/wiki/Tendrl-with-a-secure-etcd-cluster

PR #735, opened 2017-09-15
Added etcd_username and etcd_password for gluster integration conf

shtripat:
While import cluster flow, the gluster integration conf file gets
generated. Added the required etcd authentication fields here.

tendrl-bug-id: Tendrl/commons#721

### DIFF
--- a/tendrl/commons/flows/import_cluster/gluster_help.py
+++ b/tendrl/commons/flows/import_cluster/gluster_help.py
@@ -90,6 +90,8 @@ def import_gluster(parameters):
     ]['tags']['tendrl-gluster-integration']
     config_data = {"etcd_port": int(NS.config.data['etcd_port']),
                    "etcd_connection": str(NS.config.data['etcd_connection']),
+                   "etcd_username": "root",
+                   "etcd_password": str(NS.config.data['etcd_password']),
                    "log_cfg_path": (logging_config_file_path +
                                     logging_file_name),
                    "log_level": "DEBUG",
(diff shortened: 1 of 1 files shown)


shtripat:
@nthomas-redhat @r0h4n plz review

tendrl/commons/flows/import_cluster/gluster_help.py
```diff
@@ -90,6 +90,8 @@ def import_gluster(parameters):
     ]['tags']['tendrl-gluster-integration']
     config_data = {"etcd_port": int(NS.config.data['etcd_port']),
                    "etcd_connection": str(NS.config.data['etcd_connection']),
+                   "etcd_username": "root",
```
r0h4n: Why are you hardcoding root username when we have a config called etcd_username ?

tendrl/commons/flows/import_cluster/gluster_help.py
```diff
@@ -90,6 +90,8 @@ def import_gluster(parameters):
...
+                   "etcd_password": str(NS.config.data['etcd_password']),
```
r0h4n: Check if etcd_username and etcd_password actually exist in config before setting them. Don't set them if they don't exist (in case of cert file Based with)

tendrl/commons/flows/import_cluster/gluster_help.py
shtripat: Done. Will take the values from node-agent's conf and set here.

tendrl/commons/flows/import_cluster/gluster_help.py
shtripat: Ah yes. will change the code accordingly.

Issue #710, opened 2017-08-31
Import cluster job doesnt wait for other nodes import jobs to finish


shtripat:
As long as cluster details are synced in etcd and cluster is marked as `is_managed = "yes"`, the import job is marked as completed.

This might be problematic for the aggregated cluster details like rebalance, geo-replication as only onde node has reported data yet. And till other gluster-integrations start and report data, this aggregated data might be wrong.

shtripat:
@r0h4n ^^

PR #739, opened 2017-09-20
Modified to wait for import jobs to complete on other nodes

shtripat:
The parent import cluster job should be marked as finished only
when other nodes report import job as completed. Modified the code
to wait for no of nodes * 1 minutes for other jobs to complete and
then only mark the parent job as complete.

tendrl-bug-id: Tendrl/commons#710


shtripat:
@r0h4n @nthomas-redhat @nnDarshan plz review

PR #740, opened 2017-09-28
Modified the wait time per node for completing the import cluster

shtripat:
As the collectd configuration might take more time, modified the
wait time for individual node's import to complete to 2 minutes.


shtripat:
@r0h4n the wait  time of 1 minutes per node while import cluster is not enough as collectd configuration takes time on the nodes. Changed the wait time to 2 mins per nodes and executed the import cluster cluster multiple rounds. Works as expected with these changes.

Issue #749, opened 2017-10-09
Add a wrapper configure_monitoring flow around configure_monitoring atom


anmolbabu:
This is required for https://github.com/Tendrl/node-agent/issues/641

anmolbabu:
Closing this issue as the PR to fix this is merged

PR #750, opened 2017-10-09
Add configure_monitoring

anmolbabu:
This is to handle cases of provisioner disconnects and
old provisioner back in life.

tendrl-bug-id: Tendrl/commons#749

### DIFF
--- a/tendrl/commons/flows/configure_monitoring/__init__.py
+++ b/tendrl/commons/flows/configure_monitoring/__init__.py
@@ -1,0 +1,16 @@
+from tendrl.commons import flows
+from tendrl.commons.flows.exceptions import FlowExecutionFailedError
+
+
+class ConfigureMonitoring(flows.BaseFlow):
+    def __init__(self, *args, **kwargs):
+        super(ConfigureMonitoring, self).__init__(*args, **kwargs)
+
+    def run(self):
+        _cluster = NS.tendrl.objects.Cluster(
+            integration_id=NS.tendrl_context.integration_id
+        ).load()
+        if _cluster.is_managed != "yes":
+            raise FlowExecutionFailedError('Cluster is not managed')
+        self.parameters['Service.name'] = 'collectd'
+        super(ConfigureMonitoring, self).run()
--- a/tendrl/commons/objects/definition/master.yaml
+++ b/tendrl/commons/objects/definition/master.yaml
@@ -21,6 +21,24 @@ namespace.tendrl:
     ceph-provisioner: "provisioner/ceph"
     ceph-installer: "provisioner/ceph"
   flows:
+    ConfigureMonitoring:
+      tags:
+        - "tendrl/node_$NodeContext.node_id"
+      atoms:
+        - tendrl.objects.Cluster.atoms.ConfigureMonitoring
+      help: "(Re)Configure Monitoring"
+      enabled: true
+      inputs:
+        mandatory:
+          - TendrlContext.integration_id
+      pre_run:
+        - tendrl.objects.Cluster.atoms.CheckClusterAvailable
+      post_run:
+        - tendrl.objects.Service.atoms.CheckServiceStatus
+      run: tendrl.flows.ConfigureMonitoring
+      type: Create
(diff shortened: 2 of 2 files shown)


shtripat:
@anmolbabu where is the logic to invoke this flow? Link with the same.

anmolbabu:
@shtripat , @r0h4n offered to send that pr(invocation part)
>@anmolbabu where is the logic to invoke this flow? Link with the same.

tendrl/commons/objects/definition/master.yaml
```diff
@@ -21,6 +21,20 @@ namespace.tendrl:
     ceph-provisioner: "provisioner/ceph"
     ceph-installer: "provisioner/ceph"
   flows:
+    ConfigureMonitoring:
+      tags:
+        - "tendrl/node_$NodeContext.node_id"
```
anmolbabu: It is essential to make sure that the node is part of a cluster to be able to execute this flow.
So anyother tag required here?
certainly tag with provisioner contained in it might not be applicable as there is nothing that restricts this flow to be executed only on a provisioner node..

@r0h4n @shtripat

shtripat: @anmolbabu you may opt for `tendrl/integration/{node_id}` as this is added by `gluster-integration` for the node. @r0h4n ?

r0h4n: tendrl/node_$id  is a valid and accurate tag. `tendrl/integration/$node_id` should be used to target jobs to gluster-integration (job type "sds")

tendrl/commons/objects/definition/master.yaml
```diff
@@ -21,6 +21,20 @@ namespace.tendrl:
...
+      atoms:
+        - tendrl.objects.Cluster.atoms.ConfigureMonitoring
+      help: "(Re)Configure Monitoring"
+      enabled: true
+      inputs:
+        mandatory:
+          - "Node[]"
```
r0h4n: Why do you need input "Node[]"?

tendrl/commons/flows/configure_monitoring/__init__.py
```diff
@@ -0,0 +1,9 @@
+from tendrl.commons import flows
+
+
+class ConfigureMonitoring(flows.BaseFlow):
+    def __init__(self, *args, **kwargs):
+        super(ConfigureMonitoring, self).__init__(*args, **kwargs)
+
+    def run(self):
```
r0h4n: add before calling super
_cluster = NS.tendrl.objects.Cluster(integration_id=NS.tendrl_context.integration_id).load()
if _cluster.is_managed != "yes": then  raise FlowExecutionFailedError(appropriate msg here)

tendrl/commons/objects/definition/master.yaml
anmolbabu: That's the one that's used to target the flow to a particular node right?
Or the tags suffices?

tendrl/commons/flows/configure_monitoring/__init__.py
anmolbabu: ack

tendrl/commons/objects/definition/master.yaml
```diff
@@ -21,6 +21,17 @@ namespace.tendrl:
     ceph-provisioner: "provisioner/ceph"
     ceph-installer: "provisioner/ceph"
   flows:
+    ConfigureMonitoring:
```
anmolbabu: I have removed the "inputs" part as I need no "inputs" for this flow.
Is it ok?

r0h4n: Assuming ConfigureMonitoring atom doesnt require any inputs,You still need whatever inputs that required by below atom

https://github.com/Tendrl/commons/blob/master/tendrl/commons/objects/cluster/atoms/check_cluster_available/__init__.py

anmolbabu: Ack

PR #762, opened 2017-10-17
Check ansible runner output for error

shtripat:
While installation of gluster-integration, modified the logic to
check for any errors reported by ansible and populate it back to
the job status.

tendrl-bug-id: Tendrl/node-agent#627


shtripat:

PR #768, opened 2017-10-31
Use FQDN rather in logs while import cluster flow

shtripat:


shtripat:
@r0h4n @nnDarshan @nthomas-redhat plz review.
Done in context of https://github.com/Tendrl/node-agent/issues/627#issuecomment-340509498

PR #772, opened 2017-11-09
Enable tendrl-gluster-integration service during import

nnDarshan:
Currently tendrl-gluster-integration service is not
enabled during import. Because of this it doesnt start
automatically on reboot. This patch adds code to enable
the service during import.

tendrl-bug-id: Tendrl/gluster-integration#294
Signed-off-by: Darshan N <darshan.n.2024@gmail.com>

### DIFF
--- a/tendrl/commons/flows/import_cluster/gluster_help.py
+++ b/tendrl/commons/flows/import_cluster/gluster_help.py
@@ -146,7 +146,12 @@ def import_gluster(parameters):
         )
     )
     os.chmod(_gluster_integration_conf_file_path, 0o640)
+
+    if NS.config.data['package_source_type'] == 'rpm':
+        command = "systemctl enable tendrl-gluster-integration"
+        subprocess.Popen(command.split())
     subprocess.Popen(_cmd.split())
+
     return True
 
 


nnDarshan:
@nthomas-redhat @shtripat @r0h4n Please review

r0h4n:
@nnDarshan thanks. What other Tendrl services need such fix?

nnDarshan:
> @nnDarshan thanks. What other Tendrl services need such fix?

Others would not need any such fix as those are started by tendrl-ansible and its ansible's responsibility to enable the service. AFAIK this is the only service we are starting from code

tendrl/commons/flows/import_cluster/gluster_help.py
```diff
@@ -146,7 +146,12 @@ def import_gluster(parameters):
         )
     )
     os.chmod(_gluster_integration_conf_file_path, 0o640)
+
+    if NS.config.data['package_source_type'] == 'rpm':
+        command = "systemctl enable tendrl-gluster-integration"
+        subprocess.Popen(command.split())
```
r0h4n: please handle case where command fails and log appropriate error

nnDarshan: ack

PR #822, opened 2018-02-13
Renamed cluster_id attribute from message object to integration_id

GowthamShanmugam:
also modified to log the mesages using log_utils function

tendrl-bug-id: Tendrl/commons#814


tendrl/commons/flows/__init__.py
```diff
@@ -4,8 +4,8 @@
 from tendrl.commons.event import Event
```
shtripat: Do you need `Event`?

tendrl/commons/flows/create_cluster/ceph_help.py
```diff
@@ -4,67 +4,53 @@
```
r0h4n: remove this file and any other ceph file

tendrl/commons/logger.py
```diff
@@ -54,16 +54,16 @@ def push_operation(self):
...
+                self.message.integration_id != ""):
                 NS.node_agent.objects.ClusterMessage(
                     message_id=self.message.message_id,
                     timestamp=self.message.timestamp,
                     priority=self.message.priority,
                     publisher=self.message.publisher,
                     node_id=self.message.node_id,
                     payload=self.message.payload,
-                    cluster_id=self.message.cluster_id,
+                    integration_id=self.message.integration_id,
```
r0h4n: any impact on UI/API?

tendrl/commons/objects/cluster/atoms/create/__init__.py
```diff
@@ -5,9 +5,9 @@
 from tendrl.commons.flows.create_cluster.gluster_help import create_gluster
```
r0h4n: we no longer support create cluster, please remove this file and supporting files

GowthamShanmugam: i will remove it

tendrl/commons/logger.py
GowthamShanmugam: no, it won't affect UI/API

tendrl/commons/flows/create_cluster/ceph_help.py
GowthamShanmugam: i will remove it

tendrl/commons/flows/__init__.py
GowthamShanmugam: yes, because ExceptionMessage class is using Event, for that we can't use log_util

GowthamShanmugam: in this file ExceptionMessage is used so i am keeping Event

shtripat: ok

r0h4n:
@GowthamShanmugam please fix failing tests due to ceph removals

PR #826, opened 2018-02-20
Added fields volume_profiling_flag and volume_profiling_state

shtripat:
Added the fields volume_profiling_flag and volume_profiling_state
for cluster object

tendrl-bug-id: Tendrl/gluster-integration#553

### DIFF
--- a/tendrl/commons/flows/import_cluster/__init__.py
+++ b/tendrl/commons/flows/import_cluster/__init__.py
@@ -49,8 +49,8 @@ class ImportCluster(flows.BaseFlow):
                 _cluster = NS.tendrl.objects.Cluster(
                     integration_id=NS.tendrl_context.integration_id
                 ).load()
-                _cluster.enable_volume_profiling = self.parameters[
-                    'Cluster.enable_volume_profiling']
+                _cluster.volume_profiling_flag = self.parameters[
+                    'Cluster.volume_profiling_flag']
                 _cluster.save()
 
                 # Try to claim "provisioner/:integration_id" tag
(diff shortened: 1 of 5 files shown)


shtripat:
@a2batic @nthomas-redhat @r0h4n @GowthamShanmugam @gnehapk plz review

GowthamShanmugam:
LGTM

r0h4n:
Could you please fix the vulture report too https://travis-ci.org/Tendrl/commons/jobs/343844783

@shtripat

r0h4n:
@shtripat In the vulture report https://travis-ci.org/Tendrl/commons/jobs/344273070

It clearly says "tendrl/commons/__init__.py:562: unused variable 'importer' "  , which means you need to either use the variable or remove it since it is not used.

Same for "unused function" and "unused attribute" , just remove the function or attribute if it is not really used much.

shtripat:
@r0h4n so these kind issues could be a lot in projects and fixing all of them as part this PR, not sure how feasible or good that is. For new sources I can take care at the moment on this PR but for all I feel we should do a separate task only for these kind of cleanups. What you say?

r0h4n:
@shtripat  Ack https://github.com/Tendrl/commons/issues/808

tox.ini
```diff
@@ -41,6 +41,11 @@ deps =
     vulture
 commands = vulture tendrl
 
+[testenv:prospector]
```
r0h4n: please make a new PR for this along with an issue. Cant take this in this PR

shtripat: Done

Issue #805, opened 2018-01-11
tendrl need to support expand cluster functionality for gluster cluster


shtripat:
Expand cluster mechanism currently in tendrl is not very seamless.
User needs to un-manage the cluster and then extend the cluster by adding new peers. After this the cluster needs to be imported.

Ideally tendrl should be able to figure out new peers introduced in the cluster while cluster is being managed by tendrl. User should be prompted to accept the newly discovered host for monitoring / management through tendrl.

shtripat:
@r0h4n @julienlim @mcarrano @nthomas-redhat @brainfunked 
@r0h4n add you comments

julienlim:
@shtripat @r0h4n @nthomas-redhat @brainfunked @mcarrano @Tendrl/tendrl-qe 

With regards to expand cluster, why does a user have to accept newly discovered hosts for monitoring?  If new volumes or bricks get added, does that mean user has to accept them too?

IMHO, cluster expansion or changes should automatically be detected without user having to do anything -- specifically, user should not have to unmanage the cluster, then reimport the cluster, in order for the new nodes/peers to be recognized.

Is there a way for automatic expansion in the Tendrl UI without the user having to unmanage and reimport the cluster?  Meaning, all User does is add the peers/nodes to the Gluster cluster, then User adds the Tendrl agents onto the new nodes, and starts/enables the agents, generates relevant dashboards for the nodes.  Tendrl detects and adds a notification about new nodes detected and now added for management under Tendrl.  Those nodes should then automatically show up in cluster details, hosts list (for the cluster), etc. without further user intervention.

shtripat:
@julienlim my bad for using the word `accept`. Rather I meant was that user should be notified and he just acks the new nodes/peer in the cluster. Even today the volumes, brick additions etc are taken care automatically.

Regarding automatic expansion in tendrl, yes its not there at the moment and this issue is for tracking changes for the same feature.

shtripat:
Based on latest discussion below would be the flow for automatic expansion of the cluster
1. User installs glusterfs on the fresh node and uses peer probe to add the node to the existing gluster cluster
2. Using tendrl-ansible, the tendrl-node-agent gets installed on the storage node
3. Step-2 makes sure node is known to tendrl now, as part usual sync, tendrl figures out that there is a new node in the cluster and starts the import cluster flow on the new node (only) to setup required components like collectd and tendrl-gluster-integration 

Finally the new node is part of the new cluster now.

[...]

PR #835, opened 2018-02-27
Added event log for newly added nodes to the cluster

shtripat:
tendrl-bug-id: Tendrl/commons#805


shtripat:
@r0h4n @nthomas-redhat @GowthamShanmugam plz review

Issue #837, opened 2018-02-27
Attributes of Tendrl objects should be watchable


r0h4n:
- Add "watch_attrs" (type: list) to Tendrl definition schema for the Object
   - eg: 
```
objects: 
    Cluster:
      enabled: True
      help: "Represents a cluster"
      list: /clusters
      value: clusters/$TendrlContext.integration_id
      watch_attrs: 
              - errors
              - status
```

- In Services (tendrl-node-agent, tendrl-gluster-integration etc) that can .save() an object, a watcher thread will be created per attribute of that instance of the object.

eg: /clusters/asdadwwasd/status  (creates 1 watcher per service for given integration_id/status)

r0h4n:
Target is to generate status change alerts for Volumes, Clusters, Bricks, Nodes via the object.watch_attrs functionality. 

@shtripat  @Tendrl/admins any other suggestions?

shtripat:
Agree as that would be best place to do this.

PR #848, opened 2018-03-04
Attributes of Tendrl objects should be watchable

r0h4n:
tendrl-bug-id: tendrl/commons#837

### DIFF
--- a/tendrl/commons/objects/__init__.py
+++ b/tendrl/commons/objects/__init__.py
@@ -28,6 +28,7 @@ def thread_safe(thread_unsafe_method):
 @six.add_metaclass(abc.ABCMeta)
 class BaseObject(object):
     def __init__(self, *args, **kwargs):
+        self._ttl = None
         self._lock = threading.RLock()
         # Tendrl internal objects should populate their own self._defs
         if not hasattr(self, "internal"):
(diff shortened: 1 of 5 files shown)


r0h4n:
@shtripat @nthomas-redhat  We need to implement "on_change" for any object that wants to act on changes to its watched attributes (watch_attrs list in defs)

shtripat:
@r0h4n It would be great if you implement one case say for volume status. That would become a reference for all other cases.

tendrl/commons/objects/__init__.py
```diff
@@ -28,6 +28,7 @@ def thread_safe_method(self, *args, **kws):
 @six.add_metaclass(abc.ABCMeta)
 class BaseObject(object):
     def __init__(self, *args, **kwargs):
+        self._ttl = None
```
shtripat: Wouldn't it affect all other objects?

r0h4n: The variable names are different, all objects use "ttl", its only now we are storing the "ttl" passed to save as self._ttl  (there's an underscore).

So this wont affect any object except the NodeContext for now

shtripat: ok. otherwise looks fine to me

Issue #850, opened 2018-03-06
Load_all function return list with empty object when directory is empty in etcd


GowthamShanmugam:
If any directory is empty then load_all function should return empty list, but it returning list with empty object

GowthamShanmugam:
This issue is fixed, i am closing this

PR #851, opened 2018-03-06
Directory is empty then load_all function should return empty list

GowthamShanmugam:
tendrl-bug-id: Tendrl/commons#850
tendrl-bug-id: Tendrl/gluster-integration#576

Signed-off-by: GowthamShanmugamsundaram <gshanmug@redhat.com>

### DIFF
--- a/tendrl/commons/objects/__init__.py
+++ b/tendrl/commons/objects/__init__.py
@@ -169,8 +169,14 @@ class BaseObject(object):
                         return None
         ins = []
         for item in etcd_resp.leaves:
-            self.value = item.key
-            ins.append(self.load())
+            # When directory is not empty then NS._int.client.read(key)
+            # will return key + directory id as new key. If directory is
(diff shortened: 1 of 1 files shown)


GowthamShanmugam:
@shtripat @r0h4n please review

tendrl/commons/objects/__init__.py
```diff
@@ -169,8 +169,10 @@ def load_all(self):
                         return None
         ins = []
         for item in etcd_resp.leaves:
-            self.value = item.key
-            ins.append(self.load())
+            if item.key.strip("/") != value.strip("/"):
```
shtripat: what is this checking of key and value part? Can you add an example here?

GowthamShanmugam: i will add

GowthamShanmugam:
Tested

Issue #860, opened 2018-03-09
Fail node lock acquiring while import cluster only if locking job is in progress


shtripat:
if the earlier import job has failed and still nodes are locked by that job, we can safely overwrite the lock if import flow is tried again for the cluster with same set of nodes. This locking should fail only if job is in progress.

PR #861, opened 2018-03-10
Fail node locking only if job is in progress in import flow

shtripat:
tendrl-bug-id: Tendrl/commons#860

### DIFF
--- a/tendrl/commons/flows/utils.py
+++ b/tendrl/commons/flows/utils.py
@@ -166,11 +166,16 @@ def acquire_node_lock(parameters):
             if p_job_id == lock_owner_job:
                 continue
             else:
-                raise FlowExecutionFailedError("Cannot proceed further, "
-                                               "Node (%s) is already locked "
-                                               "by Job (%s)" % (node,
-                                                                lock_owner_job)
-                                               )
+                job = NS.tendrl.objects.Job(job_id=lock_owner_job).load()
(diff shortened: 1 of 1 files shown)


shtripat:
DONT MERGE. NEED TO VERIFY.

shtripat:
@GowthamShanmugam can we verify this in the setup you are having for expand flow testing?

GowthamShanmugam:
we are releasing all nodes in finally block when any exception raised in import cluster atom https://github.com/Tendrl/commons/blob/master/tendrl/commons/objects/cluster/atoms/import_cluster/__init__.py#L175

why we need this change?

GowthamShanmugam:
if we calling release lock from finally then we no need to worry about release lock

shtripat:
If everything is happy path. still we should fail node locking only if parent job is in progress. If its failed/finished node locking should be allowed. @r0h4n comments?

r0h4n:
I dont see the need to add this PR. Is it for some specific race condition between failure/finish and node lock release? @shtripat

shtripat:
I remember I ran in this issue once where import job was failed and I immediately tried import again.

r0h4n:
@shtripat  then you need to check why the nodes werent unlocked when the import job failed

shtripat:
Abandoning at the moment. and would think later if we see this issue

Issue #868, opened 2018-03-15
Add new object NS.tendrl.objects.Index


r0h4n:
Tendrl uses indexes via etcd keys, currently all Indexes are managed via etcd write/reads.

These indexes need to be encapsulated as part of a generic `NS.tendrl.objects.Index` object so that it can inherit all the goodness of the Tendrl BaseObject and make it simpler to manage all indexes

PR #871, opened 2018-03-15
tendrl-bug-id: tendrl/commons#868

r0h4n:


r0h4n:
@shtripat  @GowthamShanmugam  @Tendrl/admins please review

r0h4n:
@Tendrl/admins @shtripat @GowthamShanmugam  no reviews?

tendrl/commons/objects/definition/master.yaml
```diff
@@ -1165,4 +1165,14 @@ namespace.tendrl:
       value: /clusters/{0}/nodes/{1}/alert_counters
       list: /clusters/{0}/nodes/{1}/alert_counters
       help: "Cluster Alert Counter"
+    Index:
+      enabled: True
+      attrs:
+        key:
```
GowthamShanmugam: Here I saw key is an attribute for Index but in Index class attributes key is not used as a member variable, any reason for that?

GowthamShanmugam: Is it not necessary to present as a member variable?

tendrl/commons/objects/index/__init__.py
```diff
@@ -0,0 +1,58 @@
...
+            try:
+                curr_nodes = json.loads(curr_nodes)
+            except (ValueError, TypeError):
+                # No existing nodes have this tag index
+                pass
+
+            if NS.node_context.node_id in curr_nodes:
+                self._is_stale = False
+            else:
+                if "provisioner" in key:
```
GowthamShanmugam: I feel,  every time if we read provisioner then provisioner node will change

GowthamShanmugam: This is not an issue any way we have a mechanism to change provisioner node

GowthamShanmugam: only one major doubt i have here looks like all attributes are private except value, we are not allowing to store private arrtibutes.

tendrl/commons/objects/index/__init__.py
```diff
@@ -0,0 +1,58 @@
...
+
+class Index(objects.BaseObject):
+
+    def __init__(self, key=None, index_value=None,
+                 *args, **kwargs):
+        super(Index, self).__init__(*args, **kwargs)
+        self._index_name = key.strip("/").rsplit("/", 1)[-1]
+        self.value = "indexes/{0}".format(key.strip("/").rsplit("/", 1)[0])
+        self._is_stale = True
+        self._index_value = index_value
```
GowthamShanmugam: only one major doubt i have here looks like all attributes are private except value, we are not allowing to store private arrtibutes.

tendrl/commons/objects/index/__init__.py
shtripat: @GowthamShanmugam but changing provisioner every time there is a read to find out who provisioner would cause issues as we have seen that in history.

r0h4n:
Closing this PR for now, making a Tendrl object for indexes needs changes to Tendrl BaseObject (need compare and write etc)

Issue #867, opened 2018-03-14
Node_agent alerts are not handled by alert handlers


GowthamShanmugam:
Other components log messages are pushed into the node_agent socket and message handler in node-gent is processing the log messages. If the log message is in notice priority then it is passed to alert handler otherwise it is pushed into the logger. but if any log message which is pushed by node-agent then it won't go to the node-agent socket, it is directly passed to the logger. So the alert message which is raised from node-agent is not processed.

PR #872, opened 2018-03-15
Node_agent log message with notice priority is pushed into node_agent socket

GowthamShanmugam:
tendrl-bug-id: Tendrl/commons#867

### DIFF
--- a/tendrl/commons/objects/definition/master.yaml
+++ b/tendrl/commons/objects/definition/master.yaml
@@ -820,6 +820,7 @@ namespace.tendrl:
           - afr_quorum_state
           - afr_subvol_state
           - georep_status
+          - cluster_status
       enabled: true
       value: alerting/clusters/$Alert.tags['integration_id']/$Alert.alert_id
       list: alerting/clusters/$Alert.tags['integration_id']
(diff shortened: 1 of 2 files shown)


GowthamShanmugam:
Tendrl/node-agent/pull/742

GowthamShanmugam:
@shtripat @r0h4n please review

GowthamShanmugam:
node_agent log messages are directly pushed into logger, but node_agent message handler only getting alert from the socket and pushed into the alert handler. So in current implementation alert, log message which has notice priority also pushed into logger only, not into message handler so the alert message does not hand over to the message handler. Soi changed logic like any log_message which came from node-agent and if it has notice priority then it will be pushed into the socket. other messages pushed into logger directly.

tendrl/commons/objects/definition/master.yaml
```diff
@@ -820,6 +820,7 @@ namespace.tendrl:
           - afr_quorum_state
           - afr_subvol_state
           - georep_status
+          - cluster_status
```
r0h4n: why have you added this?

GowthamShanmugam: https://github.com/Tendrl/node-agent/pull/742/files#diff-3dfeadf899a1fe4e98d3f57feef0b4daR84

@shtripat is added new alert called cluster_status for raising new peer node identify logic,
alert messages should comes under at least any one classification. otherwise alert count won't be increased

PR #876, opened 2018-03-16
Set current job status properly when import job timed out

shtripat:
tendrl-bug-id: Tendrl/commons#875


shtripat:
@r0h4n plz review

PR #884, opened 2018-03-23
Added logic to remove current node from nodes list for expansion

shtripat:
The expand cluster job is targeted at provisioner node which for
sure would be participating in cluster by the time expand cluster
gets executed. Remove the current node from the nodes list while
figuring out where all to execute import cluster for new nodes.


shtripat:
verified

Issue #886, opened 2018-03-28
Any exception or error in load_all it return empty array


GowthamShanmugam:
Currently, if any etcd key is not found then load_all is returning None value, If load_all is used in any for loop then it will raise None type is not an iterable exception. So any error or exception in load_all then should return empty array.

PR #887, opened 2018-03-28
Any exception or error in load_all it return empty array

GowthamShanmugam:
tendrl-bug-id: Tendrl/commons#886

### DIFF
--- a/tendrl/commons/objects/__init__.py
+++ b/tendrl/commons/objects/__init__.py
@@ -172,7 +172,7 @@ class BaseObject(object):
                     # if dir is empty then item.key and value is same
                     self.value = item.key
                     ins.append(self.load())
-                    time.sleep(1)
+                    time.sleep(0.3)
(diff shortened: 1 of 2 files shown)


GowthamShanmugam:
@r0h4n @shtripat please review

tendrl/commons/objects/__init__.py
```diff
@@ -167,8 +168,7 @@ def load_all(self):
                         NS._int.reconnect()
                         etcd_resp = NS._int.client.read(value)
                     else:
-                        return None
```
r0h4n: Please use etcd_utils.read instead of NS._int.client.read, also you dont need to handle etcd.EtcdConnectionFailed, etcd.EtcdException, just  "etcd.EtcdKeyNotFound" should be handled

GowthamShanmugam:
@r0h4n changes done

Issue #799, opened 2018-01-09
Remove import statements for tendrl objects


r0h4n:
Use NS.$namespace.objects.$objectname instead of  importing the objects via python import

Tracking this bug for all the other Tendrl repos

r0h4n:
@GowthamShanmugam please take this up after M2

GowthamShanmugam:
Ok i will fix this

PR #892, opened 2018-04-03
Removing import statements for tendrl objects

GowthamShanmugam:
tendrl-bug-id: Tendrl/commons#799


GowthamShanmugam:
Please don't merge, it need to be tested, I am doing changes on other modules also

Issue #895, opened 2018-04-07
Un-manage cluster should be allowed even if previous import has failed


shtripat:
The un-managing of cluster should be allowed even if previous import has failed and un-managed has not completed on the cluster.

The check IsClusterManaged as pre-check while un-manage cluster is not required keeping the above in mind. Also while un-manage cluster flow we should check if a job is in progress for cluster which is of type import/un-manage/expand then only fail un-manage cluster flow else continue.

shtripat:
Closing as the PR is merged upstream master already

PR #896, opened 2018-04-07
Allow un-manage cluster even if there is a failed import cluster earlier

shtripat:
tendrl-bug-id: Tendrl/commons#895


shtripat:
@nthomas-redhat plz review as its a critical fix.

r0h4n:
@shtripat fix pep8

PR #771, opened 2017-11-09
Added brick_utilization also in node alert count

GowthamShanmugam:
tendrl-bugid: Tendrl/commons#770

Signed-off-by: GowthamShanmugam <gshanmug@redhat.com>

PR #901, opened 2018-04-12
Not set any default fqdn for node

r0h4n:
tendrl-bug-id: Tendrl/commons#771


GowthamShanmugam:
LGTM

Issue #923, opened 2018-04-19
Import cluster flow fails after unmanage flow


GowthamShanmugam:
During unmanage cluster_node_context was not cleared, problem is is_managed flag in cluster_node_context is always "yes". So nodes are rejected for import cluster

GowthamShanmugam:
This issue is fixed

PR #924, opened 2018-04-19
Fixing a issue import cluster fail after unmanage

GowthamShanmugam:
tendrl-bug-id: Tendrl/commons#923


GowthamShanmugam:

GowthamShanmugam:
i have tried more than 4 times import and unmanage flow, it is working fine without any issue

GowthamShanmugam:
Problem is cluster_node_context have is_managed flag True, so the node is not considered for next import

Issue #930, opened 2018-04-23
Job thread is consumes lot of cpu and memory


GowthamShanmugam:
Job thread is processing all jobs at each sync. Even if the job is finished it won't omit.

PR #931, opened 2018-04-23
Job thread in all tendrl components consumes more CPU and memory

GowthamShanmugam:
tendrl-bug-id: Tendrl/commons#930


GowthamShanmugam:
tested

Issue #949, opened 2018-04-30
After unmanage flow cluster related alerts are not deleted properly


GowthamShanmugam:
Un-managed cluster related are displayed in alerts window in UI.

PR #950, opened 2018-04-30
Removing all alerts which is related to un-managed cluster

GowthamShanmugam:
tendrl-bug-id: Tendrl/commons#949
bugzilla: 1573110


GowthamShanmugam:
@r0h4n @nthomas-redhat please review

Issue #951, opened 2018-04-30
Sometimes import cluster fails with another job in progress


GowthamShanmugam:
Shortname redundancy check and short name length check conditions fail then the job is marked as failed but status in the cluster is not modified. So status value is always in_progress, it won't allow any other jobs to continue.

GowthamShanmugam:
I am closing this issue, for now, if I reproduce then I will open it again

PR #952, opened 2018-04-30
Sometimes import cluster fails with another job in progress

GowthamShanmugam:
tendrl-bug-id: Tendrl/commons#951


GowthamShanmugam:
This Pr is invalid

PR #954, opened 2018-05-02
Import cluster job fails for a while but then finishes successfully

GowthamShanmugam:
tendrl-bug-id: Tendrl/commons/issues/953
bugzilla: 1571244


GowthamShanmugam:
@r0h4n @nthomas-redhat please review

Issue #963, opened 2018-05-08
unmanage fails without descriptive error message


GowthamShanmugam:
During un-manage delete monitoring job is not pushing any job updates.

PR #964, opened 2018-05-08
Unmanage fails without descriptive error message

GowthamShanmugam:
tendrl-bug-id: Tendrl/commons#963
bugzilla: 1565978


GowthamShanmugam:
If any exception happens in delete monitoring flow then child job for delete monitoring data is marked as failed but parent job is always waiting for child job status and timeout.

The child job is not created with parent_job_id so job_updates are not tracked by a parent.

Issue #967, opened 2018-05-14
Import not working. Job remains in "new" state


anmolsachan:
This is my latest setup on Vagrant 
:
 tendrl-api-httpd-1.6.3-20180504T083214.dd2b661.noarch
tendrl-monitoring-integration-1.6.3-20180509T112234.38a5cc7.noarch
tendrl-selinux-1.5.4-20180227T085901.984600c.noarch
tendrl-ui-1.6.3-20180510T101244.4921b67.noarch
tendrl-notifier-1.6.3-20180420T054953.dca64f3.noarch
tendrl-commons-1.6.3-20180510T100155.fb2fb63.noarch
tendrl-api-1.6.3-20180504T083214.dd2b661.noarch
tendrl-grafana-selinux-1.5.4-20180227T085901.984600c.noarch
tendrl-node-agent-1.6.3-20180504T075650.787c4d1.noarch
tendrl-grafana-plugins-1.6.3-20180509T112234.38a5cc7.noarch

anmolsachan:
@r0h4n  Can you please check this.

r0h4n:
@anmolsachan  You and @GowthamShanmugam  need to work on all backend issues, 
 
Please let me know if you guys are blocked while fixing this

PR #969, opened 2018-05-14
 Fixed Import not working. Job remains in "new" state

anmolsachan:
tendrl-bug-id: #967 


GowthamShanmugam:
LGTM

Issue #819, opened 2018-02-06
Retention policy for tendrl data


shtripat:
We need to have a retention policies for data maintained within tendrl for graphite and etcd

1. In case of managed cluster how long we should retain graphite data
2. In case of un-managed cluster, what happens to the archived data? Whether we get back the old data when the cluster is managed again or we start having graphite from the time its managed back again
3. How long we keep the alerts and notification data in tendrl central store? Currently we have some default values set for them.
4. If some document impact (need to be documented) when and where to look for archive graphite data. Currently we have plans for having a configurable path where graphite data gets archived in case of un-manage cluster and the same is communicated when a cluster is un-managed in a notification.
5. If we retain the same cluster-id (integration-id) across un-manage and manage back of cluster

Above are few questions which we need clarifications. Tagging required people for the same.

@julienlim @r0h4n @nthomas-redhat @jjkabrown1

julienlim:
@shtripat Here are some of my thoughts to the 5 points above:

(1) & (3) Regarding how long to retain data or rather what should be the Tendrl default retention policy, I've tried to list down what are default retention periods for other products/projects:

* [ManageIQ](https://access.redhat.com/documentation/en-us/reference_architectures/2017/html/deploying_cloudforms_at_scale/capacity_and_utilization#data_retention): 6 months for capacity and utilization data
[Microsoft Azure Application Insights](https://docs.microsoft.com/en-us/azure/application-insights/app-insights-data-retention-privacy): 90 days
* [OpenShift](https://docs.openshift.com/container-platform/3.6/scaling_performance/scaling_cluster_metrics.html): 7 days for Cluster metrics; samples every 15 seconds
* [oVirt](https://www.ovirt.org/documentation/data-warehouse/Application_Settings_for_the_Data_Warehouse_service_in_ovirt-engine-dwhd.conf/):  
Hourly data stored for 60 days and Daily data stored for 5 years; samples every 60 seconds)

Note: for different industries, there may be regulatory rules around retention, e.g. for PCI compliance (PCI DSS Requirement 10.7), audit logs must be kept for 1 year with a minimum of 3 months immediately available for analysis.

We will need to offer configuration for different time periods of the different types of data since there may be different rules/policies around it, and here's my initial thoughts:
* telemetry data (metrics): 60 days or 90 days
* events log (aka notification data): 6 months
* alerts: 2 days (as-is)

[...]

PR #976, opened 2018-05-23
Added a flag delete_telemetry_data for flow UnmanageCluster

shtripat:
Based on this flag passed with value `yes`, the monitoring data
maintained for the cluster under carbon would be deleted. If flag
is not passed or passed as `no`, the data would be archived in
the pre-defined path.

bugzilla: 1582375
tendrl-bug-id: Tendrl/commons#819


shtripat:
@r0h4n There is no specific BZ for removal of graphite data while un-manage of the cluster. This point came up as part of discussion under https://github.com/Tendrl/commons/issues/819#issuecomment-367886647.

Issue #977, opened 2018-05-25
unmanaged task always fails after import failure


GowthamShanmugam:
Steps to Reproduce:
1. Change tendrl-node-devel repository baseurl to bad one
2. Start cluster import
3. Import fails
4. Start cluster unmanage task

GowthamShanmugam:
This issue is fixed

PR #978, opened 2018-05-25
unmanaged task always fails after import failure

GowthamShanmugam:
tendrl-bug-id: Tendrl/commons#977
bugzilla: 1570048

### DIFF
--- a/tendrl/commons/objects/cluster/atoms/import_cluster/__init__.py
+++ b/tendrl/commons/objects/cluster/atoms/import_cluster/__init__.py
@@ -186,15 +186,15 @@ class ImportCluster(objects.BaseAtom):
                         )
                         return False
                     time.sleep(10)
-                    finished = True
+                    completed = True
                     for child_job_id in parent_job.children:
                         child_job = NS.tendrl.objects.Job(
                             job_id=child_job_id
                         ).load()
-                        if child_job.status != "finished":
-                            finished = False
+                        if child_job.status not in ["finished", "failed"]:
+                            completed = False
                             break
-                    if finished:
+                    if completed:
                         break
                     else:
                         loop_count += 1


tendrl/commons/objects/cluster/atoms/import_cluster/__init__.py
```diff
@@ -186,20 +186,38 @@ def run(self):
...
-                        if child_job.status != "finished":
-                            finished = False
-                            break
-                    if finished:
+                        if child_job.status == "finished":
+                            child_job_finished_count += 1
+                        if child_job.status == "failed":
+                            child_job_finished_count += 1
+                            child_job_failed.append(child_job.job_id)
+                    if child_job_finished_count == len(node_list) - 1:
```
GowthamShanmugam: earlier case if anyone child job fails then it is marking parent job as failed but the parent is checking only child finished or not it is not checking child is failed, so parent job is waiting until timeout.

So the user may be thought a job is failed and let's go with unmanage, but unmanage flow will fail because one storage node is still stuck with import job. 

here i am checking finished node + failed node count if it is equal to a number of child node then i am stopping job.

tendrl/commons/objects/job/__init__.py
```diff
@@ -33,17 +33,6 @@ def save(self, ttl=None):
             _parent.children = list(set(_children))
             _parent.save()
 
-        if self.status == "failed":
-            if "parent" in self.payload:
-                _parent = Job(job_id=self.payload['parent']).load()
-                _msg = "\n Child job %s failed" % self.job_id
-                if _parent.errors:
```
GowthamShanmugam: this is moved to parent import atom

GowthamShanmugam:

tendrl/commons/objects/job/__init__.py
shtripat: Can you refer the commons change here?

tendrl/commons/objects/cluster/atoms/import_cluster/__init__.py
```diff
@@ -186,20 +186,38 @@ def run(self):
...
                             job_id=child_job_id
                         ).load()
-                        if child_job.status != "finished":
-                            finished = False
-                            break
-                    if finished:
+                        if child_job.status == "finished":
+                            child_job_finished_count += 1
+                        if child_job.status == "failed":
+                            child_job_finished_count += 1
```
shtripat: child_job_finished_count += 1 ??

GowthamShanmugam: @shtripat i am adding total number of child_jobs completed, which means even job is failed that is also i am considering as completed.

So if all child are completed then i am comparing with the total number of child job is equal to completed then i am stoping parent job wait

GowthamShanmugam: i will mofdify child job child_job_finished_count to child_job_completed_count

GowthamShanmugam: problem is if any child job is failed then parent is wating for timeout.  using this logic, it will stop waiting imediatly when all jobs are completed with finshed or failed

tendrl/commons/objects/job/__init__.py
GowthamShanmugam: https://github.com/Tendrl/commons/pull/978/files#diff-5ef679f35c2d083d057b92f291c0f8e6R208

tendrl/commons/objects/cluster/atoms/import_cluster/__init__.py
shtripat: Ack

tendrl/commons/objects/job/__init__.py
```diff
@@ -33,17 +33,6 @@ def save(self, ttl=None):
             _parent.children = list(set(_children))
             _parent.save()
 
-        if self.status == "failed":
```
r0h4n: This will affect all jobs/flows other than "import_cluster", what about those jobs and their failures?

GowthamShanmugam:

Issue #979, opened 2018-05-29
Tendrl Dashboard showing wrong info and alert details.


anmolsachan:
- Tendrl doesn't reflect that all gluster nodes are down.
- Sometimes shows :Alerts: Node None is DOWN

Tendrl needs to show correct info about a node if it is up or down. Also correct notifications should be shown when nodes go up or down

PR #980, opened 2018-05-29
Updated logic for alert notifications and node status watch

anmolsachan:
tendrl-bug-id: #979 
bugzilla: 1519201, 1583264, 1580385

### DIFF
--- a/tendrl/commons/objects/node_context/__init__.py
+++ b/tendrl/commons/objects/node_context/__init__.py
@@ -122,6 +122,25 @@ class NodeContext(objects.BaseObject):
                 )
                 cluster_node_context.save()
                 del cluster_node_context
+                global_details = NS.tendrl.objects.GlobalDetails(
+                    integration_id=_tc.integration_id).load()
+                if global_details.status.lower() == "healthy":
+                    global_details.status = "unhealthy"
+                    global_details.save()
+                    _cluster = NS.tendrl.objects.Cluster(
+                        integration_id=_tc.integration_id
+                    ).load()
+                    msg = "Cluster:%s is %s" % (
+                        _cluster.short_name, "unhealthy")
+                    instance = "cluster_%s" % _tc.integration_id
+                    event_utils.emit_event(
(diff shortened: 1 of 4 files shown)


tendrl/commons/utils/central_store/utils.py
```diff
@@ -92,6 +92,7 @@ def watch(obj, key):
 
     while True:
         try:
+            obj = obj.load()
```
GowthamShanmugam: i feel doing load in all sync is very expensive one, so move this load logic under this line https://github.com/Tendrl/commons/pull/980/files#diff-cc2ad52a1d4152ad76a49b57cecfd59bR106 , because we are not using object anywhere except this block, so we can load object before on change call only.
and please verify once with that change

tendrl/commons/objects/node_context/__init__.py
```diff
@@ -111,10 +111,16 @@ def on_change(self, attr, prev_value, current_value):
...
-
                 _tc = NS.tendrl.objects.TendrlContext(
                     node_id=self.node_id
                 ).load()
+                # Load cluster_node_context will load node_context
+                # and it will be updated with latest values
+                cluster_node_context = NS.tendrl.objects.ClusterNodeContext(
+                    node_id=self.node_id,
+                    integration_id=_tc.integration_id
+                )
```
r0h4n: Every time on_change is called you will save a new cluster_node_context?  Please "load" this object instead of making a new one every time

cluster_node_context = NS.tendrl.objects.ClusterNodeContext(
                    node_id=self.node_id,
                    integration_id=_tc.integration_id
        ).load()

anmolsachan: Load cannot be done , because we need to update the cluster node context which internally loads node context. If we do a load , the cluster node context's few fields get overridden by the old fields which fails the purpose of this change. What we can do is , create the object -> Save the object(Details populate to etcd) -> Delete the object.

Issue #983, opened 2018-06-05
Unmanage and import cluster flow fails


GowthamShanmugam:
This does not fail always but it can be reprodusable by the following step:
   1. Start tendrl-node-agent in server and storage nodes
   2. After few minutes stop node-agent in a server and few storage-node.
   3. Wait for 150 to 200 sec to ttl delete status watcher field from node_context.
   4. Start node-agent on a server and storage nodes (which are all the nodes node-agent was stopped).
   5. Start import flow, after few minutes import will fail
   6. Start unmanage flow, after few minutes unmanage also  will fail (timeout)

Because watcher status field is deleted by ttl, and server also down so it won't mark that node as down. So when you start node-agent first node_context.save() in manger try to update but there is no change in an object so nothing will happen, watcher status field also not updated. After node-agent sync will try to save node_context with TTL. here also no change in object so save won't happen, But save method in a node_context object will try to refresh then watcher status field. So it will raise EtcdKeyNotFound Exception. 

Becasue of this problem some nodes are not pickup the job.

GowthamShanmugam:
This issue is fixed

PR #984, opened 2018-06-05
Fixing issue with import and unmanage flow fail

GowthamShanmugam:
bugzilla: 1570048
tendrl-bug-id: Tendrl/commons#983


tendrl/commons/objects/__init__.py
```diff
@@ -83,11 +83,13 @@ def load_definition(self):
 
     @thread_safe
     def save(self, update=True, ttl=None):
+        hash_key_changed = True
         if "Message" not in self.__class__.__name__:
             # If local object.hash is equal to
             # central_store object.hash, return
             if self.hash_compare_with_central_store(ttl=ttl):
-                return
```
GowthamShanmugam: If there is no change in hash keys then we are actually not updating object. But we have to update the watcher fields. Because if any watcher filed have TTL like node_context then it will be deleted when there is no change in an object. Because object like node_context fields won't change frequently. When components like node-agent or gluster-integration start if we do load and save objects then no changes in old objects fields and local objects then watcher attributes won't be created anymore, It will change when any one field in object is changed. So in this new logic object won't be updated but watcher fields are updated and if there is no watcher thread then it will create new one.

GowthamShanmugam: objects like node_context, bricks and volumes we are actually doing load when node-agent or gluster-integration is restarted.

GowthamShanmugam:

tendrl/commons/objects/__init__.py
```diff
@@ -83,11 +83,13 @@ def load_definition(self):
...
+                # No change in hashkey
+                hash_key_changed = False
```
shtripat: How this is different from old code. Earlier if there was no change in hash we were returning. Now also with new code you don't perform anything if there is no change in hash.

GowthamShanmugam: In old code watcher logic are not called it won't execute further after step 90, but if you see the watcher steps are written below this. 

So this flag will won't allow writing "data" field in etcd. But watcher is updated.

shtripat: oh ok. ack

Issue #988, opened 2018-06-14
Update cluster details when all nodes are goes down


GowthamShanmugam:
fix below scenarios:

* All nodes in Hosts page are Down 
* There remains at least one (the last shut down node) as Up in Grafana.
 * Not all bricks are Down in UI and in Grafana.

GowthamShanmugam:
this issue fixed

PR #989, opened 2018-06-14
Updating cluster details when all nodes are goes down

GowthamShanmugam:
tendrl-bug-id: Tendrl/commons#988
bugzilla: 1519201

### DIFF
--- a/tendrl/commons/objects/node_context/__init__.py
+++ b/tendrl/commons/objects/node_context/__init__.py
@@ -101,7 +101,7 @@ class NodeContext(objects.BaseObject):
                 pass
 
     def on_change(self, attr, prev_value, current_value):
-        if attr == "status":
+        if attr == "status" and "tendrl/monitor" in NS.node_context.tags:
             _tc = NS.tendrl.objects.TendrlContext(
                 node_id=self.node_id
             ).load()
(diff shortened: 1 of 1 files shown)


GowthamShanmugam:

tendrl/commons/objects/node_context/__init__.py
```diff
@@ -120,31 +120,23 @@ def on_change(self, attr, prev_value, current_value):
...
-                        "unhealthy",
-                        msg,
-                        instance,
-                        'WARNING',
-                        integration_id=_tc.integration_id
+                _cnc_new = \
+                    NS.tendrl.objects.ClusterNodeContext(
+                        node_id=self.node_id,
+                        integration_id=_tc.integration_id,
+                        first_sync_done=_cnc_old.first_sync_done,
```
shtripat: Why these fields `first_sync_done` and `is_managed` here?

tendrl/commons/objects/node_context/__init__.py
```diff
@@ -202,3 +194,95 @@ def on_change(self, attr, prev_value, current_value):
...
+        try:
+            nodes = etcd_utils.read(
+                "/clusters/%s/nodes" % integration_id
+            )
+            for node in nodes.leaves:
+                status = etcd_utils.read(
+                    "%s/NodeContext/status" % node.key
+                ).value
+                # Verify all nodes in a cluster are down
+                if status.lower() != "down":
```
shtripat: If sny of the nodes up return. right?

GowthamShanmugam: yes, one node up it won't execute further

tendrl/commons/objects/node_context/__init__.py
GowthamShanmugam: inside ClusterNodeContext we are actually loading node_context_obj and populating the values into ClusterNodeContext, but first_sync_done and is_managed values are not come from node_context for i am loading ClusterNodeContext and  while creating _cnc_new it will populate other fields from node_context and these two i am passing from loaded _cnc_old.

if i save directly without passing these two then other fields have a value from node_context but these two fields will come None.

shtripat: ok. this is a workaround as such till we keep populating ClusterNodeContext from NodeContext

shtripat:
Looks fine

Issue #993, opened 2018-06-18
Job sync thread will fail after /queue directory becomes empty


GowthamShanmugam:
Each job in /queue directory has TTL for two days. When all the jobs in a /queue directory are deleted by TTL then sync thread in all tendrl components are failing with error:

Jun 18 15:31:34 tendrl-server tendrl-monitoring-integration[6703]: Traceback (most recent call last):
Jun 18 15:31:34 tendrl-server tendrl-monitoring-integration[6703]: File "/usr/lib64/python2.7/threading.py", line 812, in __bootstrap_inner
Jun 18 15:31:34 tendrl-server tendrl-monitoring-integration[6703]: self.run()
Jun 18 15:31:34 tendrl-server tendrl-monitoring-integration[6703]: File "/usr/lib64/python2.7/threading.py", line 765, in run
Jun 18 15:31:34 tendrl-server tendrl-monitoring-integration[6703]: self.__target(*self.__args, **self.__kwargs)
Jun 18 15:31:34 tendrl-server tendrl-monitoring-integration[6703]: File "/usr/lib/python2.7/site-packages/tendrl/commons/jobs/__init__.py", line 84, in process_job
Jun 18 15:31:34 tendrl-server tendrl-monitoring-integration[6703]: job.save()
Jun 18 15:31:34 tendrl-server tendrl-monitoring-integration[6703]: File "/usr/lib/python2.7/site-packages/tendrl/commons/objects/job/__init__.py", line 23, in save
Jun 18 15:31:34 tendrl-server tendrl-monitoring-integration[6703]: if "parent" in self.payload:
Jun 18 15:31:34 tendrl-server tendrl-monitoring-integration[6703]: TypeError: argument of type 'NoneType' is not iterable

GowthamShanmugam:
In this case, the empty directory is different and directory not found is different, In fresh machine /queue directory not present, But that time everything is working fine. Problem is when directory present but it contains not jobs.

GowthamShanmugam:
Tendrl/api/issues/433

GowthamShanmugam:
This issue is fixed

PR #994, opened 2018-06-18
Job sync thread fails when /queue directory becomes empty

GowthamShanmugam:
tendrl-bug-id: Tendrl/commons#993
bugzilla: 1592487


GowthamShanmugam:
This PR requires Tendrl/api/pull/434

GowthamShanmugam:
@shtripat @nthomas-redhat @r0h4n @anmolsachan please review

GowthamShanmugam:
yes, we can use job.locked_by instead of reading watcher locked_by separately from etcd. I will modify this.

nthomas-redhat:
@GowthamShanmugam , Fix the Codacy issues before the merge

anmolsachan:
LGTM

tendrl/commons/tests/jobs/test_jobs_init.py
```diff
@@ -1,259 +1,126 @@
...
+                obj.run()
+
+
+@patch.object(logger, "log")
+def test_process_job_fail(log):
+    # no tag match
+    test_init.init()
+    NS.type = "node"
+    NS.publisher_id = "pytest"
+    NS.tendrl.objects.Job.load = MagicMock()
```
anmolsachan: Should we use MagicMock @r0h4n ?

GowthamShanmugam: same case here 

using mock/patch we can mock the function of a class which is existing one, but here I have created a mock function for a class in the name load.

Because TendrlContext class does not have load function, only BaseObject class have load function, when we create an object for TendrlContext then only base class (Baseobject) properties should be accessible by child class Obj. But here I want to create one mock function for class TendrlContext before object creation. So child class method overrides base class.

anmolsachan: ack

Issue #995, opened 2018-06-19
Node up alert requires integration_id from tendrl_context


GowthamShanmugam:
When managed node is up from down state then server machine will 
raise alert for node up, but it reqiredintegration id to raise
alert. problem is when node is up tendrl_context object is empty 
for sometime in etcd.

GowthamShanmugam:
this issue is fixed

PR #996, opened 2018-06-19
Node up alert requires integration_id from tendrl_context

GowthamShanmugam:
when managed node is up from down state then server machine will
raise alert for node up, but it reqiredintegration id to raise
alert. problem is when node is up tendrl_context object is empty
for sometime in etcd.

tendrl-bug-id: Tendrl/commons#995
bugzilla: 1580385

### DIFF
--- a/tendrl/commons/__init__.py
+++ b/tendrl/commons/__init__.py
@@ -314,7 +314,7 @@ class TendrlNS(object):
                        {"message": "Setup TendrlContext for namespace."
                                    "%s" % self.ns_name})
             self.current_ns.tendrl_context = \
-                self.current_ns.objects.TendrlContext()
+                self.current_ns.objects.TendrlContext().load()
             NS.tendrl_context = self.current_ns.tendrl_context
 
     def _create_ns(self):
--- a/tendrl/commons/tests/test_init.py
+++ b/tendrl/commons/tests/test_init.py
@@ -14,10 +14,12 @@ import yaml
 
 from tendrl.commons import objects
 import tendrl.commons.objects.node_context as node
+import tendrl.commons.objects.tendrl_context as tendrl_context
 from tendrl.commons import TendrlNS
 from tendrl.commons.utils import etcd_utils
 
 
+@patch.object(tendrl_context.TendrlContext, "load")
 @patch.object(etcd, "Client")
 @patch.object(Client, "read")
 @patch.object(node.NodeContext, '_get_node_id')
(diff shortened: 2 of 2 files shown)


GowthamShanmugam:
@shtripat @r0h4n @nthomas-redhat @anmolsachan  please review

GowthamShanmugam:
When I verified this problem in my vagrant setup this issue does not happen, but I have verified with @nthomas-redhat's high configured machine this problem I can reproduced often

tendrl/commons/__init__.py
```diff
@@ -314,7 +314,7 @@ def setup_common_objects(self):
                        {"message": "Setup TendrlContext for namespace."
                                    "%s" % self.ns_name})
             self.current_ns.tendrl_context = \
-                self.current_ns.objects.TendrlContext()
+                self.current_ns.objects.TendrlContext().load()
```
shtripat: So this makes sure we load the latest value from etcd. But how are we making sure, we raise alert only when `integration_id` is available?

GowthamShanmugam: all storage nodes should have integration_id when it is detected, so integration_id in tendrl_context should be available always. 

I feel we are not overriding or modifying tendrl_context anywhere except this place.

what my point is integtation_id is always available nowhere it is modified except this place in tendrl.

GowthamShanmugam: And @shtripat when we restart node-agent and gluster integration back to back if some time integration_id not present then leads to lot of problem

shtripat: ack

GowthamShanmugam:
Verified

tendrl/commons/tests/test_init.py
```diff
@@ -461,6 +465,7 @@ def test_setup_common_objects(monkeypatch):
         tendrlNS.current_ns.objects["Config"] = obj_cls[1]
     with patch.object(etcd, "Client", return_value=etcd.Client()) as client:
         tendrlNS.current_ns.objects.pop("NodeContext")
+        NS.tendrl.objects.TendrlContext.load = MagicMock()
```
anmolsachan: Should we use MagicMock ? Most of the places we have been using Mock/patch. Asking just consistency.

tendrl/commons/__init__.py
shtripat: So this load would resolve that issue of integration_id not present for sometime as well?

tendrl/commons/tests/test_init.py
GowthamShanmugam: using mock/patch we can mock the function of a class which is existing one, but here I have created a mock function for a class in the name load. 

Because TendrlContext class does not have load function, only BaseObject class have load function, when we create an object for TendrlContext then only base class (Baseobject) properties should be accessible by child class Obj. But here I want to create one mock function for class TendrlContext before object creation. So child class method overrides base class.

tendrl/commons/__init__.py
GowthamShanmugam: @shtripat yes it will make sure always integration_id present in tendrl_context object

tendrl/commons/tests/test_init.py
anmolsachan: ack

Issue #997, opened 2018-06-19
Format of data returned by AnsibleRunner.run() method is not consistent


anmolsachan:
Exceptions data is not returned properly in AnsibleRunner.run() method (https://github.com/Tendrl/commons/blob/36458f15dcd7d06257f244aba84a3dbeb6d13ccb/tendrl/commons/utils/ansible_module_runner.py#L114) due to which the code execution fails when there is an exception during Ansible command run.

This issue is not completely reproducible by needs to be handled.

More info:

  Failure in Job 4a5268f9-54e1-4d76-9738-af1b096e1825 Flow tendrl.objects.Node.flows.StopServices with error:
  Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/tendrl/commons/jobs/__init__.py", line 242,
    in process_job the_flow.run()
  File "/usr/lib/python2.7/site-packages/tendrl/commons/objects/node/flows/stop_services/__init__.py", line 14,
    in run srv = NS.tendrl.objects.Service(service=service)
  File "/usr/lib/python2.7/site-packages/tendrl/commons/objects/service/__init__.py", line 9,
    in __init__ service_detail = self.get_service_info(service)
  File "/usr/lib/python2.7/site-packages/tendrl/commons/objects/service/__init__.py", line 18,
    in get_service_info return {"exists": service.exists(), "running": service.status()}
  File "/usr/lib/python2.7/site-packages/tendrl/commons/utils/service_status.py", line 32,
    in status stdout, stderr, rc = self._execute_service_command('is-active')
  File "/usr/lib/python2.7/site-packages/tendrl/commons/utils/service_status.py", line 16,
    in _execute_service_command return command.run()
  File "/usr/lib/python2.7/site-packages/tendrl/commons/utils/cmd_utils.py", line 78,
    in run stdout = result.get("stdout", "") AttributeError: 'str' object has no attribute 'get' 
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

PR #998, opened 2018-06-19
Fixed inconsistency in AnsibleRunner return data

anmolsachan:
tendrl-bug-id: Tendrl/commons#997
bugzilla: 1589729

### DIFF
--- a/tendrl/commons/utils/ansible_module_runner.py
+++ b/tendrl/commons/utils/ansible_module_runner.py
@@ -111,9 +111,8 @@ class AnsibleRunner(object):
             result = json.loads(out)
 
         except (subprocess.CalledProcessError, ValueError) as ex:
-            result = repr(ex)
-            err = result
+            result = {'stderr': repr(ex)}
+            err = repr(ex)
         finally:
             os.remove(_temp_file.name)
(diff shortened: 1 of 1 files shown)


nthomas-redhat:
@anmolsachan , check why the centos build is happening

tendrl/commons/utils/ansible_module_runner.py
```diff
@@ -111,9 +111,8 @@ def run(self):
             result = json.loads(out)
 
         except (subprocess.CalledProcessError, ValueError) as ex:
-            result = repr(ex)
-            err = result
+            result = {'stderr': repr(ex)}
```
shtripat: How does it take care of `result.get('stdout')` in calling code now?

anmolsachan: Since, a dictionary is being returned now, the stdout will be assigned blank string( result.get("stdout", "") ), thus the code will not throw any error. https://github.com/Tendrl/commons/blob/36458f15dcd7d06257f244aba84a3dbeb6d13ccb/tendrl/commons/utils/cmd_utils.py#L78

PR #999, opened 2018-06-19
Logger error and ETCD_utils update 

mlanotte:


tendrl/commons/tests/test_logger.py
```diff
@@ -92,6 +98,12 @@ def test_push_event():
     message["priority"] = "error"
     log.push_event()
 
+def test_push_notification():
+    message = init()
+    log = Logger(message)
```
GowthamShanmugam: When tendrl starts each component in tendrl will create a namespace with attribute name NS with all tendrl commons objects and it own component-specific objects and flows, etcd.

For example, if you are running tendrl-node-agent then NS contains:
   NS.tendrl:  It contains all commons objects and flows
   NS.node_agent: It contains all  node-agent specific objects and flows
To test any file or function in tendrl we need to create some dummy NS with tendrl and component-specific namespace with all objects or flow or some object required by the function which you are going to test.

Here we are actually testing logger function, so it requires only ClusterMessage, NodeMessage, Message, NotifierMessage objects.

So we are creating dummy NS with some dummy objects with these name. In your case for push notification function required NotifierMessage object so got to https://github.com/mlanotte/commons/blob/b7dc00d5fbd5db5d9ad8253342504d3e82638e8e/tendrl/commons/tests/fixtures/cluster_message.py  and NotifierMessage object like

```class NotificationMessage(object):

        pass

    def save(self):
        pass
```

GowthamShanmugam: then it will work

GowthamShanmugam:
Any file_name which starts with a dot(.) those are created by the system, you should not add those files in your PR and you should not modify those files directly also. Ignore that kind of file, whenever you run tox these files will be created by the system.

mlanotte:
I did not mean to push the .pycache files. I'm fairly new to github so those were mistakes and I'm not sure how to remove them from this pull request

mbukatov:
@mlanotte no worries, since this is a pull request branch (and not a master branch), you still can fix it.

You have basically 2 options:

 * add a new commit, removing the files at once there
 * rewriting a history and removing the files from commits which added these files by mistake via interactive rebase

I see that there are only 3 commits which adds the cache files by mistake:

```
$ git log --oneline .pytest_cache/
96b9ec7 Created a test file for the events.py file that so far covers about 80% of the file
c8beae1 Created a new file to test the config.py file. Tests that an Exception is correctly raised if the test_config method is handed a non existing yaml path
b005e37 Fixed all the tests for the logger file so that it has complete coverage. Had to add a Notification field to the cluster_message file to achieve this.
```

So the rebase would be easy (ping me if you would like to see a demonstration). If you want to learn it, this is a good opportunity. Otherwise I guess that removing the files in new commit is fine. Check with @r0h4n what he prefers.

I would also suggest you to add this line:

```
.pytest_cache
```

Into `.gitignore` file of the project. This way, you won't commit these files by mistake again.

Few other notes related to git usage:

 * keep 1st line of a git commit under 70 characters, so that the git history graph is easy to oversee (unlike in the example above)
 * before you start working on something, create new git branch with a name which would suggest what the change is about, eg. `$ git checkout -b logger_etcd_updated`

mlanotte:
I'm not sure why the main init file lost 2% coverage

mlanotte:
The pull request should be all set now

tendrl/commons/tests/objects/node_context/test_node_context_init.py
```diff
@@ -179,3 +249,215 @@ def test_create_node_id(patch_etcd_utils_read,
                 with patch.object(os, "makedirs", return_value=True):
                     node_context._create_node_id()
         f.close()
+
+
+"""
+@patch.object(etcd, "Client")
```
r0h4n: why is this code commented? If not required, please remove it

tendrl/commons/tests/objects/node_context/test_node_context_init.py
```diff
@@ -179,3 +249,215 @@ def test_create_node_id(patch_etcd_utils_read,
...
+        node_context.render()
+        node_context.save()
+        node_context.save(ttl="test")
+        node_context.update_cluster_details(0)
+
+
+# Tried to cover past the if/return statement but so far am unable to
+# Figure out how to have that if fail
+"""
+@patch.object(etcd, "Client")
```
r0h4n: please remove the commented code and the personalized code comments should be added as a 

"TODO ($username) Tried to cover past the if/return statement but so far am unable to  Figure out how to have that if fail"

tendrl/commons/tests/test_events.py
```diff
@@ -0,0 +1,101 @@
...
+def init(patch_etcd_utils_read,
+         patch_get_node_id,
+         patch_read,
+         patch_client,
+         tc):
+    tc.return_value = tendrl_context.TendrlContext
+    patch_get_node_id.return_value = 1
+    patch_read.return_value = etcd.Client()
+    patch_client.return_value = etcd.Client()
+    # setattr(__builtin__, "NS", maps.NamedDict())
```
r0h4n: remove commented lines

tendrl/commons/tests/test_message.py
```diff
@@ -105,6 +106,9 @@ def test_constructor_Messsage():
     # mod = importlib.import_module("tendrl.commons.message")
     # sys.path.remove('/usr/lib64/collectd')
     # del(sys.modules["tendrl.commons.message"])
```
r0h4n: please remove the commented code lines

tendrl/commons/tests/test_message.py
```diff
@@ -105,6 +106,9 @@ def test_constructor_Messsage():
...
+    msg = Message("info", "node_context", payload={
+        "exception": "Test Message"}, caller=obj_caller)
+    # assert msg.node_id == "Test id"
```
r0h4n: remove the commented code line

Issue #1000, opened 2018-06-21
Cluster is marked as managed after import cluster job fails


GowthamShanmugam:
I have stopped monitoring-integration during import cluster flow, So import cluster flow is failed with error  ```Setting up cluster aliasnot yet complete. Timing out```  but after job failed cluster is marked as managed in etcd.

GowthamShanmugam:
This issue is fixed

PR #1001, opened 2018-06-21
Cluster is updated as managed after import cluster job fails

GowthamShanmugam:
tendrl-bug-id: Tendrl/commons#1000
bugzilla: 1593640

### DIFF
--- a/tendrl/commons/flows/import_cluster/__init__.py
+++ b/tendrl/commons/flows/import_cluster/__init__.py
@@ -98,7 +98,7 @@ class ImportCluster(flows.BaseFlow):
             # even from expand cluster flow. We should not set the
             # cluster's current job status from child jobs
             _job = NS.tendrl.objects.Job(job_id=self.job_id).load()
-            if 'parent' not in _job.payload:
+            if 'parent' not in _job.payload and _job.status != "failed":
                 _cluster = NS.tendrl.objects.Cluster(
                     integration_id=NS.tendrl_context.integration_id
                 ).load()


GowthamShanmugam:
@r0h4n @nthomas-redhat @shtripat @anmolbabu please review

tendrl/commons/flows/import_cluster/__init__.py
```diff
@@ -98,7 +98,7 @@ def run(self):
             # even from expand cluster flow. We should not set the
             # cluster's current job status from child jobs
             _job = NS.tendrl.objects.Job(job_id=self.job_id).load()
-            if 'parent' not in _job.payload:
+            if 'parent' not in _job.payload and _job.status != "failed":
```
shtripat: If job failed due to time out, shouldn't it have raised an exception for atom and it goes to L#109?

GowthamShanmugam: It will go when nodes its own job failed, but parent job won't go to that place when a child is failed.
When a child is failed the child job will mark parent as failed, but parent job is updating cluster is managed. So before mark cluster as managed the parent has to check job status like any child marked parent as failed.

shtripat: In that shouldnt we have the condition as `if 'parent' not in _job.payload and _job.status == 'finished'` ?

GowthamShanmugam: Problem is when all child's are finished successfully we are not marking parent as finished, Because parent still not completed its job. So if 'parent' not in _job.payload and _job.status == 'finished'  change won't work in success case. Because parent job is still in process, So cluster won't marked as Managed.

shtripat: Ack. Looks fine.

Issue #1002, opened 2018-06-21
Node down alert are raised when some nodes  are not managed by tendrl


GowthamShanmugam:
I have done new peer probe with existing cluster, and i have start node-agent in an unmanaged node to expand and after some time I shut down to a new node. After few minutes node down alert raised for an unmanaged node.

GowthamShanmugam:
This issue is fixed

PR #1003, opened 2018-06-21
Raising node down alert only for managed nodes

GowthamShanmugam:
tendrl-bug-id: Tendrl/commons#1002
bugzilla: 1580385

### DIFF
--- a/tendrl/commons/objects/node_context/__init__.py
+++ b/tendrl/commons/objects/node_context/__init__.py
@@ -105,7 +105,12 @@ class NodeContext(objects.BaseObject):
             _tc = NS.tendrl.objects.TendrlContext(
                 node_id=self.node_id
             ).load()
-            if current_value is None:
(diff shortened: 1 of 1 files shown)


GowthamShanmugam:
Handled a corner case here, all nodes are down and some nodes are ready to expand then current logic won't consider all nodes are down and it won't raise a brick alert. Here I modified logic like all managed nodes are down and any unmanaged node status is up, in this case also I am raising all brick are down and volume down.

if i am not checked unmanaged node then if we stop and down the unmanaged nodes before cluster import also node down alert will be raised.

GowthamShanmugam:
verified

GowthamShanmugam:
@r0h4n @shtripat 
@nthomas-redhat please review

GowthamShanmugam:
I verified @shtripat but I this is a very corner case so it not strike in my mind at last build.

PR #1005, opened 2018-06-21
Several unit tests, exploring coverage

Josh-Everett:
Some tests are redundant and needed to be cleaned up, mainly probing for coverage at this point.


Josh-Everett:
Unexpected error, will re-PR shortly

PR #1012, opened 2018-06-28
Added a couple additional test files

nathan-weinberg:


tendrl/commons/tests/objects/cluster/atoms/valid_import_cluster_params/test_valid_import_cluster_params.py
```diff
@@ -0,0 +1,24 @@
+import __builtin__
+import maps
+from tendrl.commons.objects.cluster.atoms.valid_import_cluster_params \
```
anmolsachan: Try keeping tendrl imports separate from library imports. We can have a line in between them

nathan-weinberg:
Note for reviewers: my latest commit includes two tests that test Tendrl components with two minute timeouts each; as such the build will take longer than usual but I have verified it does work on my own machine

PR #1013, opened 2018-06-28
Got objects/nodes to 100% and added small coverage to other

mlanotte:
Added small addition to other file


tendrl/commons/tests/objects/alert/test_alert_init.py
```diff
@@ -0,0 +1,128 @@
+import __builtin__
+import maps
+from mock import patch
+from tendrl.commons import objects
```
anmolsachan: Try keeping tendrl imports separate from library imports. We can have a line in between them.

mlanotte:
I added the line in between the imports

mlanotte:
Just realized me and Nathan both did the alert_init__ so I'm not sure how we should go about it. Both of our files make the coverage as high as possible

mlanotte:
Is there a way for my  tendrl/commons/tests/objects/alert/test_alert_init.py to be added but not Nathan's?
He worked on it first and I didn't realize it but his is only giving the file 94% coverage while mine is 100% so it would be better to take mine instead.

nathan-weinberg:
I will remove mine from my PR :(

Issue #1019, opened 2018-07-03
Child Jobs in Import and Unmanage not marked failed after parent job timeout.


anmolsachan:
Child jobs in Import and Unmanage remain in new state if parent job gets timed out and failed. Child jobs should be marked failed explicitly and should not be waited for ttl to mark them failed. These stale jobs can be picked up by node-agent in a scenario in which an import is fired but one of the nodes is down. After import failure if the node comes up, it will pick up those jobs and keep the node-agent busy and defer other tasks such as Unmanage which may fail them due to timeout.

PR #1020, opened 2018-07-03
Marking of child jobs failed after parent timeout

anmolsachan:
tendrl-bug-id: Tendrl/commons#1019
bugzilla: 1584095

### DIFF
--- a/tendrl/commons/objects/cluster/atoms/import_cluster/__init__.py
+++ b/tendrl/commons/objects/cluster/atoms/import_cluster/__init__.py
@@ -184,6 +184,13 @@ class ImportCluster(objects.BaseAtom):
                             job_id=self.parameters['job_id'],
                             flow_id=self.parameters['flow_id']
                         )
+                        for child_job_id in parent_job.children:
+                            child_job = NS.tendrl.objects.Job(
+                                job_id=child_job_id
+                            ).load()
+                            if child_job.status not in ["finished", "failed"]:
+                                child_job.status = "failed"
+                                child_job.save()
                         return False
                     time.sleep(10)
                     completed = True
--- a/tendrl/commons/objects/cluster/atoms/import_cluster/__init__.py
+++ b/tendrl/commons/objects/cluster/atoms/import_cluster/__init__.py
@@ -184,6 +184,11 @@ class ImportCluster(objects.BaseAtom):
                             job_id=self.parameters['job_id'],
                             flow_id=self.parameters['flow_id']
                         )
+                        # Marking child jobs as failed which did not complete
+                        # as the parent job has timed out. This has to be done
+                        # explicitly because these jobs will still be processed
+                        # by the node-agent, and will keep it busy, which might
+                        # defer the new jobs or lead to their timeout.
                         for child_job_id in parent_job.children:
                             child_job = NS.tendrl.objects.Job(
                                 job_id=child_job_id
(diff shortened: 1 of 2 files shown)


r0h4n:
@GowthamShanmugam  please review this

tendrl/commons/objects/cluster/atoms/import_cluster/__init__.py
```diff
@@ -184,6 +184,13 @@ def run(self):
                             job_id=self.parameters['job_id'],
                             flow_id=self.parameters['flow_id']
                         )
+                        for child_job_id in parent_job.children:
+                            child_job = NS.tendrl.objects.Job(
+                                job_id=child_job_id
+                            ).load()
+                            if child_job.status not in ["finished", "failed"]:
+                                child_job.status = "failed"
```
r0h4n: please add comment here explaining the status check and change

tendrl/commons/objects/cluster/atoms/stop_monitoring_services/__init__.py
```diff
@@ -72,6 +72,13 @@ def run(self):
                         job_id=self.parameters['job_id'],
                         flow_id=self.parameters['flow_id'],
                     )
+                    for child_job_id in child_job_ids:
+                        child_job = NS.tendrl.objects.Job(
+                            job_id=child_job_id
+                        ).load()
+                        if child_job.status not in ["finished", "failed"]:
+                            child_job.status = "failed"
```
r0h4n: please add comment here explaining the status check and change

tendrl/commons/objects/cluster/atoms/import_cluster/__init__.py
anmolsachan: Child jobs in Import and Unmanage remain in new state if parent job gets timed out and failed. Child jobs should be marked failed explicitly and should not be waited for ttl to mark them failed. These stale jobs are picked up by node-agent in a scenario in which an import is fired but one of the nodes is down. After import failure if the node comes up, it will pick up those jobs and keep the node-agent busy and defer other tasks such as Unmanage( if fired immediately after import failure) which may fail them due to timeout.

tendrl/commons/objects/cluster/atoms/stop_monitoring_services/__init__.py
anmolsachan: Child jobs in Import and Unmanage remain in new state if parent job gets timed out and failed. Child jobs should be marked failed explicitly and should not be waited for ttl to mark them failed. These stale jobs are picked up by node-agent in a scenario in which an import is fired but one of the nodes is down. After import failure if the node comes up, it will pick up those jobs and keep the node-agent busy and defer other tasks such as Unmanage( if fired immediately after import failure) which may fail them due to timeout.

r0h4n: I mean add code comment :)

tendrl/commons/objects/cluster/atoms/import_cluster/__init__.py
GowthamShanmugam: https://github.com/anmolsachan/commons/blob/7ff3d2ece3f92c5c963ddce0299bb41dc86c5b2f/tendrl/commons/objects/cluster/atoms/import_cluster/__init__.py#L181 move this logic inside timout block, when parent timeout then marks child job as fail

tendrl/commons/objects/cluster/atoms/stop_monitoring_services/__init__.py
GowthamShanmugam: same case here

GowthamShanmugam:
LGTM

tendrl/commons/objects/cluster/atoms/stop_monitoring_services/__init__.py
anmolsachan: okay.

PR #1021, opened 2018-07-03
Proposed changes to is_tendrl_node_managed

nathan-weinberg:
I moved the various tests the objects "run" method makes into separate functions for ease of testing - functionality should be identical.

I've included a modified unit test for the changed file in this PR.

### DIFF
--- a/tendrl/commons/objects/node/atoms/is_node_tendrl_managed/__init__.py
+++ b/tendrl/commons/objects/node/atoms/is_node_tendrl_managed/__init__.py
@@ -6,128 +6,15 @@ from tendrl.commons.utils import etcd_utils
 from tendrl.commons.utils import log_utils as logger
 
 
-def os_check(node_id):
-    try:
-        os_details = etcd_utils.read("nodes/%s/Os" % node_id)
-        if os_details.leaves is None:
-            logger.log(
-                "error",
-                NS.get("publisher_id", None),
-                {
-                    "message": "Node %s doesn't have OS details "
-                               "populated" % NS.node_context.fqdn
-                },
-                job_id=self.parameters['job_id'],
-                flow_id=self.parameters['flow_id']
-            )
-            return False
-    except etcd.EtcdKeyNotFound:
-        logger.log(
-            "error",
-            NS.get("publisher_id", None),
-            {
-                "message": "Node %s doesn't have OS details "
-                           "populated" %
-                           NS.node_context.fqdn
-            },
-            job_id=self.parameters['job_id'],
-            flow_id=self.parameters['flow_id']
-        )
+def check_resource(resource, node_id):
+    res = etcd_utils.read("nodes/%s/%s" % (node_id, resource))
+    if res.leaves is None:
         return False
-
-
-
-def cpu_check(node_id):
-    try:
-        cpu_details = etcd_utils.read("nodes/%s/Cpu" % node_id)
-        if cpu_details.leaves is None:
-            logger.log(
-                "error",
-                NS.get("publisher_id", None),
-                {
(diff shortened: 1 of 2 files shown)


anmolsachan:
@r0h4n Will do.

nathan-weinberg:
@anmolsachan @r0h4n I have fixed the merge errors, please review the changes. Once verified I will refactor my new test file in a separate commit to reflect the changes in the code.

tendrl/commons/objects/node/atoms/is_node_tendrl_managed/__init__.py
```diff
@@ -6,6 +6,127 @@
...
+        os_details = etcd_utils.read("nodes/%s/Os" % node_id)
+        if os_details.leaves is None:
+            logger.log(
+                "error",
+                NS.get("publisher_id", None),
+                {
+                    "message": "Node %s doesn't have OS details "
+                               "populated" % NS.node_context.fqdn
+                },
+                job_id=self.parameters['job_id'],
```
anmolsachan: self wont be accessible here because this function is not a method of the class. A workaround could be similar to : 

```
def check_resource(resource,node_id):
    res = etcd_utils.read("nodes/%s/%s" % node_id,resource)
    if res.leaves is None:
        return False
    return True

class IsNodeTendrlManaged(objects.BaseAtom):
        super(IsNodeTendrlManaged, self).__init__(*args, **kwargs)

    def run(self):
        node_ids = self.parameters.get('Node[]')
        if not node_ids or len(node_ids) == 0:
            raise AtomExecutionFailedError("Node[] cannot be empty")

        resources =["Os", "Memory", "Cpu" ,"Networks"]
        for node_id in node_ids:
            for resource in resources:
                try:
                    if not check_resource(node_id, resource):
                        logger.log(
                        "error",
                        NS.get("publisher_id", None),
                        {
                            "message": "Node %s doesn't have %s details "
                                       "populated" % (NS.node_context.fqdn,resource)
                        },
                        )
                        return False
                except etcd.EtcdKeyNotFound:
                    logger.log(
                        "error",
                        NS.get("publisher_id", None),
                        {
                            "message": "Node %s doesn't have %s details "
                                       "populated" % (NS.node_context.fqdn,resource)
                        },
                        )
                        return False
        return True
```
This might be much easier to test too.
Pardon any syntax errors, just wrote a pseudo-code

nathan-weinberg: LGTM - I will try it out and push changes if successful

anmolsachan:
LGTM.

anmolsachan:
@nathan-weinberg please test the code change once, to be sure enough.

nathan-weinberg:
@anmolsachan I'm not sure what you mean; are you referring to the unit test? If so I have already run and verified that it works

anmolsachan:
@nathan-weinberg  I meant the code, change. Since this PR changes some code too, it needs to be tested. Please install a latest version of the upstream setup and put this code there and check if the code is behaving as intended. Since this atom is part of the Import flow, check import flow with this code.

shtripat:
@nathan-weinberg I agree with @anmolsachan here. We should verify the import flow and its sanity for sure with the changes. Looking at the code it looks, it should work seamlessly though :)

nathan-weinberg:
@anmolsachan @shtripat No problem - will do and report back with the results :+1:

nathan-weinberg:
Verified code functionality did not change on a tendrl-vagrant VM setup on my local machine. CentOS CI has as of yet not created an RPM build but I have a local one available if anyone would like access.

Issue #1032, opened 2018-07-11
import flow fails for large cluster


GowthamShanmugam:
import flow timeout logic is not sufficient for large cluster, so import always fails with timeout message

GowthamShanmugam:
This issue is fixed

PR #1033, opened 2018-07-11
Modified import flow timeout logic based on number of volumes

GowthamShanmugam:
tendrl-bug-id: Tendrl/commons#1032
bugzilla: 1600092

### DIFF
--- a/tendrl/commons/objects/cluster/atoms/check_sync_done/__init__.py
+++ b/tendrl/commons/objects/cluster/atoms/check_sync_done/__init__.py
@@ -1,6 +1,7 @@
 import time
 
 from tendrl.commons import objects
+from tendrl.commons.utils import cmd_utils
 from tendrl.commons.utils import log_utils as logger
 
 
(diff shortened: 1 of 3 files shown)


GowthamShanmugam:
@shtripat is this sufficient?

tendrl/commons/objects/cluster/atoms/check_sync_done/__init__.py
```diff
@@ -10,12 +11,18 @@ def __init__(self, *args, **kwargs):
...
-
-        # wait for 360 sec to complete the first round of sync of
+        cmd = cmd_utils.Command('gluster volume list')
+        out, err, rc = cmd.run()
+        # default intervel is 4 min
+        wait_count = 240
+        if not err:
+            volumes = out.split("\n")
+            # 15 sec for each volume
+            wait_count = wait_count + (len(volumes)) * 15
```
shtripat: So is there are 3 volumes (say) in cluster, the actual wait time for each node would be as below

(240 + 3*15)5    # sleep time in loop is 5 seconds

So effectively the wait time would be 1425 second (~23.75 mins). I dont think you wanted this. Check the sleep time for each loop and based on that form the wait count.

So suppose you want to give 5 sec for each volume extra over actual wait time so wait count could be (72 + 1*no of vols) and thus for 3 volumes the total wait time would be (72+1*3) * 5 = 375 seconds.

The same calculation should be applicable for below case as well.

GowthamShanmugam:
Verified

PR #1036, opened 2018-07-13
Fixed umnanage failed after node failure

anmolsachan:
tendrl-bug-id: Tendrl/commons#1019
bugzilla: 1584095


tendrl/commons/jobs/__init__.py
```diff
@@ -189,7 +189,6 @@ def process_job(jid):
                 pass
             lock_info = dict(node_id=NS.node_context.node_id,
```
anmolsachan: I think keeping this as dict would be better than making it as a dict. Dictionary comparison works perfectly, it was the list that was the issue.

anmolsachan:
@r0h4n @shtripat @GowthamShanmugam Please review.

Issue #1039, opened 2018-07-19
Set the cluster status as expand_pending if expansion failed


shtripat:
The cluster should be marked as expand_pending if previous expansion failed. This is required to make sure user cleans the error with the underlying cluster and should be able to trigger expansion again from UI. At the moment there is no way to trigger expansion again if failed.

shtripat:
Closing as PR merged

PR #1043, opened 2018-07-25
Fixed test case failures

shtripat:
tendrl-bug-id: Tendrl/commons#1039


nthomas-redhat:
@shtripat , test case failures pls fix

PR #1046, opened 2018-08-02
Fixing expand re-run issue

GowthamShanmugam:
tendrl-bug-id: Tendrl/commons#1039
bugzilla: 1596655


GowthamShanmugam:

tendrl/commons/flows/utils.py
```diff
@@ -168,14 +168,23 @@ def acquire_node_lock(parameters):
                 if p_job_id == lock_owner_job:
                     continue
                 else:
-                    raise FlowExecutionFailedError(
-                        "Cannot proceed further, "
-                        "Node (%s) is already locked "
-                        "by Job (%s)" % (
-                            node_id,
-                            lock_owner_job
+                    # if lock owner job not none and it is not
```
shtripat: The comment should be like `if the locker owner job is already finished or failed, we should allow other flows to acquire the lock.`

Issue #1052, opened 2018-08-28
 Repeated Import (and Unmanage) fails: Timing out import job, Cluster data still not fully updated


GowthamShanmugam:

GowthamShanmugam:
For now, i am closing this issue

PR #1053, opened 2018-08-28
Repeated Import (and Unmanage) fails: Timing out import job, Cluster data still not fully updated

GowthamShanmugam:
tendrl-bug-id: Tendrl/commons#1052
bugzilla: 1616005

### DIFF
--- a/tendrl/commons/objects/cluster/atoms/delete_monitoring_details/__init__.py
+++ b/tendrl/commons/objects/cluster/atoms/delete_monitoring_details/__init__.py
@@ -2,6 +2,7 @@ import time
 import uuid
 
 from tendrl.commons import objects
+from tendrl.commons.utils import etcd_utils
 from tendrl.commons.utils import log_utils as logger
 
 
@@ -36,9 +37,13 @@ class DeleteMonitoringDetails(objects.BaseAtom):
             payload=payload
         ).save()
 
-        # Wait for 2 mins for the job to complete
+        # Find number jobs in a queue
+        queue = etcd_utils.read("/queue")
+        no_of_jobs = sum(1 for x in queue.leaves)
+        # Wait for 2 mins for the job to complete +
+        # 5 sec per 5 job for waiting time in a queue
         loop_count = 0
-        wait_count = 24
+        wait_count = 24 + (no_of_jobs / 5)
         while True:
             child_job_failed = False
             if loop_count >= wait_count:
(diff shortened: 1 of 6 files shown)


tendrl/commons/objects/cluster/atoms/delete_monitoring_details/__init__.py
```diff
@@ -36,9 +37,15 @@ def run(self):
             payload=payload
         ).save()
 
-        # Wait for 2 mins for the job to complete
+        # Find number jobs in a queue
+        no_of_jobs = 0
+        queue = etcd_utils.read("/queue")
+        for _ in queue.leaves:
```
shtripat: on `queue.leaves` there should be a way to get length using iterators I feel

shtripat: you can use `len(queue._children)` as that `_children` returns a list

GowthamShanmugam: _children is a private variable, is it correct to use this in our code?

tendrl/commons/objects/cluster/atoms/delete_monitoring_details/__init__.py
```diff
@@ -36,9 +37,13 @@ def run(self):
             payload=payload
         ).save()
 
-        # Wait for 2 mins for the job to complete
+        # Find number jobs in a queue
+        queue = etcd_utils.read("/queue")
+        no_of_jobs = sum(1 for x in queue.leaves)
```
anmolsachan: What jobs are we looking to wait for here (Which will be in new or processing state ) ? Ideally in import scenario , if it fails or it passes, all the jobs will be marked as failed or finished. Can you please give a scenario where this is happening.

GowthamShanmugam: we are actually giving waiting time for child job to stopping integration service and stopping monitoring service. let's take example stopping integration service: 
we are giving waiting time for child jobs like (no.of.nodes* 2) and sleeping time is 5. So if we use 2 storage nodes then 2*2 = 4 and 4*5 =20 so 10 sec per each job. So this time is only to stop the gluster-integration.  What about the job waiting time in a queue. when a job is pushed into the queue then 20 sec are started. but still, a job is waiting in a queue to process by some node. So a problem occurs when there is a huge number of jobs in a queue. Before job picked by any one of the nodes, the unmanaged is flow is marked as timed out from stopping tendrl service atom.

So as per this logic i am calculating number jobs in a queue and i am deciding waiting time

GowthamShanmugam: it is not about job state new or processing or finished. we need to know number of jobs in a queue and set waiting time of atom. Because we have to give sometime to the job thread to pickup the job. then we have to start out reqular waiting time to finish the job

anmolsachan: Understood

GowthamShanmugam:
This issue is not required for now, i am closing this

Issue #1054, opened 2018-09-11
tendrl-commons package should depend on specific minimal version of ansible


dahorak:
Accordingly to Tendrl/commons/issues/944 and discussion on [gitter](https://gitter.im/Tendrl/tendrl-devel?at=5b96a3a9e481f854a69a93c2) `tendrl-commons` depends on ansible in version 2.5 or higher:
> we are actually changed tendrl code based on ansible 2.5 releases,
> So it won't work with ansible 2.4 series. We are already have issue for
> this and fixed https://github.com/Tendrl/commons/issues/944#event-1595594115

But in [package requirements](https://github.com/Tendrl/commons/blob/master/tendrl-commons.spec#L16) it is only `ansible` without any version specified (output truncated):
```
# rpm -qR tendrl-commons 
ansible
...
```

This *"inconsistency"* between real and specified requirements leads to unexpected and hardly understandable issues like this: https://www.redhat.com/archives/tendrl-devel/2018-June/msg00008.html

More detailed description
--------------------------------
Package `ansible` in version `2.4.2.0-2.el7` is available in `extras` repository (by default available and enabled on any freshly installed CentOS machine).
Package `ansible` in version `2.6.3-1.el7` is available from `epel` repository, which have to be additionally added (manually or in our case automatically by *tendrl-ansible*'s `site.yml` playbook).

When I'll follow the *tendrl-ansible*'s installation instructions and I'll firstly install `tendrl-ansible` on freshly installed Tendrl Server machine, it will automatically install ansible-2.4 from the `extras` repository as a dependency.
As a next step, I'll launch `site.yml` playbook, which will add `epel` repository on Tendrl Server and all Gluster Storage Servers.
Then it will install `tendrl-commons` package. Since this package depends on *any* version of `ansible`, it will install latest available `ansible` (`2.6` from `epel` repository) on all Gluster Storage Servers, but didn't automatically update the existing `ansible-2.4` on Tendrl Server.

Expected state
-------------------
If `tendrl-commons` requires specific minimal version of `ansible`, it should be defined also in the package requirements, so it will be ensured, that the correct version of `ansible` is installed together with the `tendrl-commons` package.

GowthamShanmugam:
@TimothyAsir

dahorak:
[...]

PR #1055, opened 2018-09-11
specify minimal required ansible version

dahorak:
- fixes Tendrl/commons/issues/1054

Signed-off-by: Daniel Horak <dahorak@redhat.com>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -112,7 +112,7 @@ setup(
     license="LGPL-2.1+",
     zip_safe=False,
     install_requires=[
-        "ansible",
+        "ansible>=2.5",
         "psutil",
         "python-etcd",
         "python-dateutil",
--- a/tendrl-commons.spec
+++ b/tendrl-commons.spec
@@ -13,7 +13,7 @@ BuildRequires: python-mock
 BuildRequires: python-six
 BuildRequires: systemd
 
-Requires: ansible
+Requires: ansible >= 2.5
 Requires: python-maps
 Requires: python-dateutil
 Requires: python-dns


dahorak:
run tests

dahorak:
There might be some problem with the running ansible updating itself on the Tendrl server. I'll need to debug it slightly more deeply.

dahorak:
This patch might cause update of ansible package on Tendrl Server, when ansible-playbook is running (with `site.yml`), which probably cause [following error](https://ci.centos.org/job/tendrl-2-1-cluster-gluster-install/594/console):
```
TASK [tendrl-ansible.tendrl-server : Configure etcd and graphite ip addr in node-agent.conf.yaml] ***
task path: /usr/share/ansible/roles/tendrl-ansible.tendrl-server/tasks/tendrl-node-agent.yml:12
The full traceback is:
  File "/usr/lib/python2.7/site-packages/ansible/executor/task_executor.py", line 96, in run
    try:
  File "/usr/lib/python2.7/site-packages/ansible/executor/task_executor.py", line 296, in _run_loop
    if loop_var in task_vars:
  File "/usr/lib/python2.7/site-packages/ansible/executor/task_executor.py", line 528, in _execute
  File "/usr/lib/python2.7/site-packages/ansible/plugins/action/normal.py", line 43, in run
    wrap_async = self._task.async_val and not self._connection.has_native_async
AttributeError: 'Task' object has no attribute 'async_val'

fatal: [n53.dusty.ci.centos.org]: FAILED! => {
    "msg": "Unexpected failure during module execution.", 
    "stdout": ""
}
```

I'm not sure now, what is the best solution for this.

anmolsachan:
@dahorak If we update the spec file of tendrl-ansible itself ( https://github.com/Tendrl/tendrl-ansible/blob/4acae343ad4fab4f16ce330e047403374d0cb7a6/tendrl-ansible.spec#L15 ) , Wouldn't it solve the issue ?

dahorak:
As I wrote under [the related issue](https://github.com/Tendrl/commons/issues/1054#issuecomment-420205490), the main "problem" here is, that the expected *ansible* version (2.6) is in *epel* repository, which is added "automatically" when *tendrl-ansible*'s `site.yml` playbook is executed.
Based on the current documentation, when *tendrl-ansible* is installed, only the older version of *ansible* is available (from *CentOS-7 - Extras*). So we should also update the (pre)installation instructions to add *epel* repository manually before *tendrl-ansible* package is installed.

mbukatov:
@anmolsachan  both tendrl-ansible and tendrl-commons requires ansible, so both should specify the minimal version. In upstream, one can try to use tendrl-ansible on another machine or install it from git repo, and in such case, minimal version in tendrl-commons will have it's effect.

GowthamShanmugam:
@dahorak Need to specify the minimum version of ansible in setup.py also, otherwise it won't work when I do
python setup.py, install.

GowthamShanmugam:
LGMT

anmolsachan:
LGTM

shtripat:
LGTM

shtripat:
Fix the commit message, pep8 and squash in one commit.

PR #1059, opened 2018-10-05
Changed the md5 algo to sha512 for FIPS compliance

shtripat:
tendrl-bug-id: Tendrl/commons/issues/1058


nthomas-redhat:
@shtripat , Please check the UT failures

PR #1061, opened 2018-10-11
Added job for profiling after import

anmolsachan:
tendrl-bug-id: Tendrl/commons#1062
bugzilla: 1638314
Signed-off-by: Anmol Sachan anmol13694@gmail.com

### DIFF
--- a/tendrl/commons/flows/import_cluster/__init__.py
+++ b/tendrl/commons/flows/import_cluster/__init__.py
@@ -1,14 +1,17 @@
+import datetime
 import etcd
 import json
 import re
 import sys
 import traceback
+import uuid
 
 
 from tendrl.commons import flows
 from tendrl.commons.flows.exceptions import FlowExecutionFailedError
 from tendrl.commons.objects import AtomExecutionFailedError
 from tendrl.commons.utils import etcd_utils
+from tendrl.commons.utils import log_utils as logger
 
 
 class ImportCluster(flows.BaseFlow):
@@ -109,6 +112,54 @@ class ImportCluster(flows.BaseFlow):
                 _cluster.current_job['status'] = "finished"
                 _cluster.is_managed = "yes"
                 _cluster.save()
+
+                # Creating a job to enable/disable profiling based on flag
+                # selected while import flow. This job is separate from
+                # import flow and not a child flow. This could be tracked
+                # independently once import if over.
+                if _cluster.volume_profiling_flag != "leave-as-is":
+                    # Possible  cluster.volume_profiling_state are enabled,
+                    # disabled and mixed
+                    # Possible cluster.volume_profiling_flag are enable,
+                    # disable and leave-as-is.
+                    # Below we compare the given above ( state and flag)
+                    if (_cluster.volume_profiling_state !=
+                            _cluster.volume_profiling_flag + "d"):
+                        _job_id = str(uuid.uuid4())
+                        params = {"TendrlContext."
+                                  "integration_id":
+                                      NS.tendrl_context.integration_id,
+                                  "Cluster.volume_profiling_flag":
+                                      _cluster.volume_profiling_flag
+                                  }
+                        payload = {"tags": ["provisioner/%s" %
+                                            NS.tendrl_context.integration_id],
+                                   "run": "gluster.flows.EnableDisableVolume"
+                                          "Profiling",
+                                   "status": "new",
+                                   "parameters": params,
+                                   "type": "sds",
+                                   "created_at": (datetime.datetime.
+                                                  utcnow().strftime
+                                                  ('%Y-%m-%d %H:%M:%SZ')),
+                                   "created_from": "BackendFlow"
+                                   }
+                        NS.tendrl.objects.Job(
+                            job_id=_job_id,
+                            status="new",
+                            payload=payload
+                        ).save()
+                        logger.log(
+                            "info",
+                            NS.publisher_id,
+                            {"message": "Created job ( %s ) for profiling. "
+                                        "The job could be tracked independent"
+                                        "ly once this import job is completed"
+                                        "." % _job_id},
+                            job_id=self.parameters['job_id'],
+                            flow_id=self.parameters['flow_id']
+                        )
+
         except (FlowExecutionFailedError,
                 AtomExecutionFailedError,
                 Exception) as ex:


anmolsachan:
@shtripat @nthomas-redhat @GowthamShanmugam Please review.

tendrl/commons/flows/import_cluster/__init__.py
```diff
@@ -109,6 +112,46 @@ def run(self):
                 _cluster.current_job['status'] = "finished"
                 _cluster.is_managed = "yes"
                 _cluster.save()
+
+                # Creating a Job to do the profiling task. This job is
```
shtripat: `Creating a job to enable/disable profiling based on flag selected while import flow. This job is separate from import flow and not a child flow. This could be tracked independently once import if over.`

Does this make more sense?

tendrl/commons/flows/import_cluster/__init__.py
```diff
@@ -109,6 +112,46 @@ def run(self):
...
+                #  separate from the import flow.
+                if _cluster.volume_profiling_flag != "leave-as-is":
+                    if (_cluster.volume_profiling_state !=
+                            _cluster.volume_profiling_flag+"d"):
```
shtripat: We should add some comment here explaining why we are adding `d` at the last to check `_cluster.volume_profiling_state`. I feel its because valid values of `_cluster.volume_profiling_state` are `Enabled`, `Disabled` and `Mixed`.
Please add comment to explain the same.

tendrl/commons/flows/import_cluster/__init__.py
```diff
@@ -109,6 +112,46 @@ def run(self):
...
+                        _job_id = str(uuid.uuid4())
+                        params = {"TendrlContext.integration_id":
```
shtripat: Shouldnt we pass the whole import cluster flow's parameters here?

tendrl/commons/flows/import_cluster/__init__.py
```diff
@@ -109,6 +112,46 @@ def run(self):
...
+                                            NS.tendrl_context.integration_id],
+                                   "run": "gluster.flows.EnableDisableVolume"
+                                          "Profiling",
+                                   "status": "new",
+                                   "parameters": params,
+                                   "type": "sds",
+                                   "created_at": (datetime.datetime.
+                                                  utcnow().strftime
+                                                  ('%Y-%m-%d %H:%M:%SZ')),
+                                   "created_from": "BackendSpecialCase"
```
shtripat: `backend_flow`

tendrl/commons/flows/import_cluster/__init__.py
```diff
@@ -109,6 +112,46 @@ def run(self):
...
+                                   }
+                        NS.tendrl.objects.Job(
+                            job_id=_job_id,
+                            status="new",
+                            payload=payload
+                        ).save()
+                        logger.log(
+                            "info",
+                            NS.publisher_id,
+                            {"message": "Created job ( %s ) for profiling. "
```
shtripat: `Created job (%s) for enable/disable profiling. The job could be tracked independently once this import job is completed.`
Does this look better?

anmolsachan: yes, Will do the change

tendrl/commons/flows/import_cluster/__init__.py
```diff
@@ -109,6 +112,54 @@ def run(self):
...
+                # import flow and not a child flow. This could be tracked
+                # independently once import if over.
+                if _cluster.volume_profiling_flag != "leave-as-is":
+                    # Possible  cluster.volume_profiling_state are enabled,
+                    # disabled and mixed
+                    # Possible cluster.volume_profiling_flag are enable,
+                    # disable and leave-as-is.
+                    # Below we compare the given above ( state and flag)
+                    if (_cluster.volume_profiling_state !=
+                            _cluster.volume_profiling_flag + "d"):
```
GowthamShanmugam: i think we can modify this condition like "if _cluster.volume_profiling_flag in _cluster.volume_profiling_state" or create a map {"enabled": enable, "disabled": "disable"} and use the map here. appendind "d" and end i feel not good.

anmolsachan: I thought of doing the similar earlier, but then I realized that for just one comparison it would complicate the logic. I added the comments to understand what is happening. Is it fine ?

GowthamShanmugam:
Profiling enabling logic is blocking cluster-integration sync, It will take huge time for the cluster which has a large number of volumes. So by this time volumes details may be deleted by TTL.

@shtripat @cloudbehl we need some logic to handle this here, otherwise no use of reducing import time.

GowthamShanmugam:
Profiling enabling logic is blocking cluster-integration sync, It will take huge time for the cluster which has a large number of volumes. So by this time volumes details may be deleted by TTL.

@shtripat @cloudbehl we need some logic to handle this here, otherwise no use of reducing import time.

GowthamShanmugam:
Profiling enabling logic is blocking cluster-integration sync, It will take huge time for the cluster which has a large number of volumes. So by this time volumes details may be deleted by TTL.

@shtripat @cloudbehl we need some logic to handle this here, otherwise no use of reducing import time.

GowthamShanmugam:
Profiling enabling logic is blocking cluster-integration sync, It will take huge time for the cluster which has a large number of volumes. So by this time volumes details may be deleted by TTL.

@shtripat @cloudbehl we need some logic to handle this here, otherwise no use of reducing import time.

tendrl/commons/flows/import_cluster/__init__.py
GowthamShanmugam: Profiling enabling logic is blocking cluster-integration sync, It will take huge time for the cluster which has a large number of volumes. So by this time volumes details may be deleted by TTL.

GowthamShanmugam:
Profiling enabling logic is blocking cluster-integration sync, It will take huge time for the cluster which has a large number of volumes. So by this time volumes details may be deleted by TTL.

@shtripat @cloudbehl we need some logic to handle this here, otherwise no use of reducing import time.

tendrl/commons/flows/import_cluster/__init__.py
GowthamShanmugam: Profiling enabling logic is blocking cluster-integration sync, It will take huge time for the cluster which has a large number of volumes. So by this time volumes details may be deleted by TTL.

@shtripat @cloudbehl we need some logic to handle this here, otherwise no use of reducing import time.

anmolsachan:
Closing this as a better way was found. Decided to keep profiling in the beginning of the import, within the flow.

Issue #1074, opened 2019-01-11
In upstream code import is failing when no volumes present in the cluster


GowthamShanmugam:

GowthamShanmugam:
this issue is fixed

PR #1075, opened 2019-01-11
Import is failing with new profiling code when no volumes present in the cluster

GowthamShanmugam:
tendrl-bug-id: Tendrl/commons#1074
bugzilla: 1676897

### DIFF
--- a/tendrl/commons/objects/cluster/atoms/import_cluster/__init__.py
+++ b/tendrl/commons/objects/cluster/atoms/import_cluster/__init__.py
@@ -58,14 +58,15 @@ class ImportCluster(objects.BaseAtom):
                         return False
                 else:
                     logger.log(
-                        "error",
+                        "warning",
                         NS.publisher_id,
-                        {"message": "Failed to fetch volumes info for "
-                                    "enable/disable profiling."},
+                        {"message": "Unable to find list of volumes in a "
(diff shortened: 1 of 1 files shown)


GowthamShanmugam:
@shtripat @nthomas-redhat @anmolsachan please review

tendrl/commons/objects/cluster/atoms/import_cluster/__init__.py
```diff
@@ -56,6 +56,16 @@ def run(self):
                             flow_id=self.parameters['flow_id']
                         )
                         return False
+                elif err == "No volumes present in cluster":
```
shtripat: Looks fine but is there better way to find this error, instead of looking of string matching?

GowthamShanmugam: my point is we can log the error message, it can any error and we can continue import flow Because users can also enable/disable profiling manually.


So we just display an error message in warning and we will continue the import. We don't need to check specific message here.

shtripat: ok

GowthamShanmugam:
@shtripat @nthomas-redhat need to take this PR

GowthamShanmugam:
image

Issue #1076, opened 2019-03-13
Incorrect number of hosts down reported 


GowthamShanmugam:
Grafana sometimes reports wrong number of hosts down when all nodes are shut down.

When I stop services tendrl-node-agent, collectd and tendrl-gluster-integration then for the first time grafana usually shows correctly that all nodes are down but if I start them and after a while I stop these services again then grafana reports that 4 hosts are down and 2 are up. This is happening consistently with multiple installation with 6 nodes.

GowthamShanmugam:
this issue is fixed

GowthamShanmugam:
This issue is fixed

PR #1077, opened 2019-03-13
Incorrect number of hosts down reported

GowthamShanmugam:
tendrl-bug-id: Tendrl/commons#1076
bugzilla: 1687333

### DIFF
--- a/tendrl/commons/objects/node_context/__init__.py
+++ b/tendrl/commons/objects/node_context/__init__.py
@@ -93,6 +93,11 @@ class NodeContext(objects.BaseObject):
     def save(self, update=True, ttl=None):
         super(NodeContext, self).save(update)
         status = self.value + "/status"
+        if self.status == "UP" and ttl is None:
+            # Set ttl always when node status in up
+            ttl = int(
+                NS.config.data.get("sync_interval", 60)
+            )
         if ttl:
             self._ttl = ttl
             try:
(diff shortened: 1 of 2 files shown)


tendrl/commons/utils/central_store/utils.py
```diff
@@ -107,7 +107,8 @@ def watch(obj, key):
                     attr = key.rstrip("/").split("/")[-1]
                     obj.on_change(attr, prev_val,
                                   cur_val)
-        except etcd.EtcdKeyNotFound:
+        except Exception:
+            # When watch crash then clear key from the watchers
```
GowthamShanmugam: Sometimes watcher is crashing when a key is deleted by TTL (but not for all the time, need to figure out why),  we have to remove the key to create new watcher when save function called for that particular object.

GowthamShanmugam:
@fbalak @shtripat please review

tendrl/commons/utils/central_store/utils.py
```diff
@@ -107,7 +107,10 @@ def watch(obj, key):
                     attr = key.rstrip("/").split("/")[-1]
                     obj.on_change(attr, prev_val,
                                   cur_val)
-        except etcd.EtcdKeyNotFound:
+        except Exception as ex:
+            if type(ex) is etcd.EtcdEventIndexCleared:
+                continue
```
GowthamShanmugam: Move waiting index to latest

GowthamShanmugam:

tendrl/commons/objects/node_context/__init__.py
```diff
@@ -93,6 +93,11 @@ def render(self):
     def save(self, update=True, ttl=None):
         super(NodeContext, self).save(update)
         status = self.value + "/status"
+        if self.status == "UP" and ttl == None:
```
shtripat: `ttl is None` ?

tendrl/commons/objects/node_context/__init__.py
```diff
@@ -93,6 +93,11 @@ def render(self):
...
+            # Set ttl always when node status in up
```
shtripat: The comment says that ttl is set always if status is UP but condition above says status UP and ttl None

tendrl/commons/utils/central_store/utils.py
```diff
@@ -107,7 +107,10 @@ def watch(obj, key):
                     attr = key.rstrip("/").split("/")[-1]
                     obj.on_change(attr, prev_val,
                                   cur_val)
-        except etcd.EtcdKeyNotFound:
+        except Exception as ex:
+            if type(ex) is etcd.EtcdEventIndexCleared:
```
shtripat: Add a comment to explain why this movement of wait indexd required

tendrl/commons/objects/node_context/__init__.py
GowthamShanmugam: This check will always be assigning TTL value when node status is up, if we are saving node object with status "up" without TTL then TTL is an optional parameter in a save function, so it will check status is up and TTL None then it will assigning TTL while saving an object.

 TTL is important for the watcher function because we are marking node status is down based on TTL only. I saw sometimes node status is updated as up without TTL. So it is affecting watcher to modify status as down when a node is actually down.

We can always remember to give TTL wherever we use node object save. so i put that in a common place.

tendrl/commons/objects/node_context/__init__.py
GowthamShanmugam: we are already passing TTL while calling save method, but we can't always remember to pass TTL value if TTL value is not passed and status of the node is up then it will assign default TTL value.

we can pass any TTL value also. if TTL not there it will assign a default value.

GowthamShanmugam:
@shtripat changes done, please review

tendrl/commons/objects/node_context/__init__.py
shtripat: All I meant was instead of saying `ttl == None` we should say `ttl is None` :)

tendrl/commons/objects/node_context/__init__.py
shtripat: Ack

tendrl/commons/utils/central_store/utils.py
shtripat: Looks good now

tendrl/commons/objects/node_context/__init__.py
GowthamShanmugam: oh :)

Issue #1080, opened 2019-03-26
Import flow failure related log messages should be more specific about what went wrong


GowthamShanmugam:
In most of the error log messages in import flow are very generic, It displays a big traceback with atom failed messages. But it is not specified why the atom is failed. With this error message user unable to pinpoint of the failure. 

## import failed

Failure in Job 0845ffb1-4d53-4a6c-9e18-3ed0a72c1ce5 Flow tendrl.flows.ImportCluster with error: Traceback (most recent call last): File "/usr/lib/python2.7/site-packages/tendrl/commons/jobs/__init__.py", line 240, in process_job the_flow.run() File "/usr/lib/python2.7/site-packages/tendrl/commons/flows/import_cluster/__init__.py", line 131, in run exc_traceback) FlowExecutionFailedError: ['Traceback (most recent call last):\n', ' File "/usr/lib/python2.7/site-packages/tendrl/commons/flows/import_cluster/__init__.py", line 98, in run\n super(ImportCluster, self).run()\n', ' File "/usr/lib/python2.7/site-packages/tendrl/commons/flows/__init__.py", line 227, in run\n "Error executing post run function: %s" % atom_fqn\n', 'AtomExecutionFailedError: Atom Execution failed. Error: Error executing post run function: tendrl.objects.Cluster.atoms.SetupClusterAlias\n'

Failed post-run: tendrl.objects.Cluster.atoms.SetupClusterAlias for flow: Import existing Gluster Cluster

GowthamShanmugam:
This issue is fixed, please close this

PR #1081, opened 2019-03-26
Import flow failure related log messages should be more specific about what went wrong

GowthamShanmugam:
tendrl-bug-id: Tendrl/commons#1080
bugzilla: 1688630

### DIFF
--- a/tendrl/commons/objects/cluster/atoms/import_cluster/__init__.py
+++ b/tendrl/commons/objects/cluster/atoms/import_cluster/__init__.py
@@ -74,6 +74,7 @@ class ImportCluster(objects.BaseAtom):
                         # create same flow for each node in node list except
                         #  $this
                         payload = {"tags": ["tendrl/node_%s" % node],
+                                   "node_id": node,
                                    "run": "tendrl.flows.ImportCluster",
                                    "status": "new",
                                    "parameters": new_params,
@@ -115,7 +116,8 @@ class ImportCluster(objects.BaseAtom):
                 logger.log(
                     "error",
                     NS.publisher_id,
-                    {"message": "Failed to detect underlying cluster version"},
+                    {"message": "Failed to detect underlying cluster "
+                     "version. Error: %s" % err},
                     job_id=self.parameters['job_id'],
                     flow_id=self.parameters['flow_id']
                 )
@@ -220,15 +222,6 @@ class ImportCluster(objects.BaseAtom):
                         job_id=self.parameters['job_id']
                     ).load()
                     if loop_count >= wait_count:
-                        logger.log(
-                            "error",
-                            NS.publisher_id,
-                            {"message": "Import jobs on cluster(%s) not yet "
-                             "complete on all nodes(%s). Timing out." %
-                             (_cluster.short_name, str(node_list))},
-                            job_id=self.parameters['job_id'],
-                            flow_id=self.parameters['flow_id']
-                        )
                         # Marking child jobs as failed which did not complete
                         # as the parent job has timed out. This has to be done
                         # explicitly because these jobs will still be processed
@@ -239,10 +232,41 @@ class ImportCluster(objects.BaseAtom):
                                 job_id=child_job_id
                             ).load()
                             if child_job.status not in ["finished", "failed"]:
+                                if child_job.status in ["new", ""]:
+                                    node_id = child_job.payload.get(
+                                        "node_id", ""
+                                    )
+                                    node_obj = NS.tendrl.objects.NodeContext(
+                                        node_id=node_id
+                                    ).load()
+                                    logger.log(
+                                        "error",
+                                        NS.publisher_id,
+                                        {"message": "Import child job %s is "
+                                         "not yet picked by %s, Either node is"
+                                         " down or tendrl-node-agent service "
+                                         "is not running" % (
+                                             child_job.job_id,
+                                             node_obj.fqdn
+                                         )},
+                                        job_id=self.parameters['job_id'],
+                                        flow_id=self.parameters['flow_id']
+                                    )
                                 child_job.status = "failed"
                                 child_job.save()
+
+                        logger.log(
+                            "error",
+                            NS.publisher_id,
+                            {"message": "Import jobs on cluster(%s) not yet "
+                             "complete on all nodes(%s). Timing out." %
+                             (_cluster.short_name, str(node_list))},
+                            job_id=self.parameters['job_id'],
+                            flow_id=self.parameters['flow_id']
+                        )
+
                         return False
-                    time.sleep(10)
+                    time.sleep(5)
                     completed = True
                     for child_job_id in parent_job.children:
                         child_job = NS.tendrl.objects.Job(
--- a/tendrl/commons/objects/cluster/atoms/setup_cluster_alias/__init__.py
+++ b/tendrl/commons/objects/cluster/atoms/setup_cluster_alias/__init__.py
@@ -36,13 +36,23 @@ class SetupClusterAlias(objects.BaseAtom):
         while True:
             child_job_failed = False
             if loop_count >= wait_count:
+                job = Job(job_id=_job_id).load()
+                if job.status in ["new", ""]:
+                    msg = ("Child job %s for setting up cluster alias not yet "
+                           "picked up by the server, Service "
+                           "tendrl-monitoring-integration may be down. "
+                           "Timing out. (%s)" % (
+                               job.job_id, integration_id
+                           ))
+                else:
+                    msg = ("Child job %s for setting up cluster alias not yet"
+                           "compeleted by the server. Timing out. (%s)" % (
+                               job.job_id, integration_id))
                 logger.log(
                     "error",
                     NS.publisher_id,
                     {
-                        "message": "Setting up cluster alias"
-                        "not yet complete. Timing out. (%s)" %
-                        integration_id
+                        "message": msg
                     },
                     job_id=self.parameters['job_id'],
                     flow_id=self.parameters['flow_id'],
--- a/tendrl/commons/objects/definition/master.yaml
+++ b/tendrl/commons/objects/definition/master.yaml
@@ -68,6 +68,7 @@ namespace.tendrl:
         optional:
           - Cluster.short_name
       pre_run:
+        - tendrl.objects.Node.atoms.CheckRequiredServicesRunning
         - tendrl.objects.Cluster.atoms.CheckClusterNodesUp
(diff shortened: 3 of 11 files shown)


GowthamShanmugam:
https://github.com/Tendrl/monitoring-integration/pull/594/files

tendrl/commons/jobs/__init__.py
```diff
@@ -94,11 +95,10 @@ def process_job(jid):
...
     # tendrl-node-agent tagged as tendrl/monitor will ensure
-    # >10 min old "new" jobs are timed out and marked as
-    # "failed" (the parent job of these jobs will also be
-    # marked as "failed")
-    if "tendrl/monitor" in NS.node_context.tags and \
-        _timeout == "yes" and job.status == "new":
+    # >10 min old "new" parent jobs are timed out and marked 
+    # as "failed"
+    if "tendrl/monitor" in NS.node_context.tags and _timeout == "yes" and \
+        job.status == "new" and job.payload.get('parent') is None:
```
GowthamShanmugam: timeout only parent job, when it is not processed for a long time.  Because child job will be timed out by a parent job with the proper log message. Timing out from here not give an exact error message.

tendrl/commons/jobs/__init__.py
```diff
@@ -209,7 +198,7 @@ def process_job(jid):
                     obj_name, flow_name)
             else:
                 runnable_flow = current_ns.ns.get_flow(flow_name)
-
+            time.sleep(3)
             job = job.load()
```
GowthamShanmugam: this will avoid same job picked by different nodes

tendrl/commons/objects/cluster/atoms/import_cluster/__init__.py
```diff
@@ -239,10 +232,41 @@ def run(self):
...
+                             "complete on all nodes(%s). Timing out." %
+                             (_cluster.short_name, str(node_list))},
+                            job_id=self.parameters['job_id'],
+                            flow_id=self.parameters['flow_id']
+                        )
+
                         return False
-                    time.sleep(10)
+                    time.sleep(5)
                     completed = True
```
GowthamShanmugam: 10 sec per sleep is too much for import fail, we did all calculation for 5 sec only, it will work for a large number of volumes. because we are setting timeout based on a number of volumes. Also, we reduced gluster-integration sync time so we don't need 10 sec sleep.

tendrl/commons/objects/node/flows/stop_services/__init__.py
```diff
@@ -23,50 +24,64 @@ def run(self):
...
+                        job_id=self.parameters['job_id'],
+                        flow_id=self.parameters['flow_id'],
+                    )
                 continue
 
             _cmd_str = "systemctl stop %s" % service
             cmd = cmd_utils.Command(_cmd_str)
-            err, out, rc = cmd.run()
+            out, err, rc = cmd.run()
             if err:
```
GowthamShanmugam: https://github.com/Tendrl/commons/blob/master/tendrl/commons/utils/cmd_utils.py#L81 cmd utils return out, err rc. While disabling service it always raises error related removed "removed symlink", That is why we previously gave like that. But that is not the correct approach, so now i have modified log level to debug to avoid this in log file.

tendrl/commons/jobs/__init__.py
```diff
@@ -314,7 +303,7 @@ def process_job(jid):
                        "job status invalid"
                 raise FlowExecutionFailedError(_msg)
             else:
-                job = job.load()
```
GowthamShanmugam: Sometimes job.load() is not updating object with new values

GowthamShanmugam:

tendrl/commons/jobs/__init__.py
shtripat: This will need a exhaustive testing of multiple jobs.

tendrl/commons/objects/cluster/atoms/import_cluster/__init__.py
shtripat: Actually as per the calculation it should be sleep for 5 sec only so that 96*5 = 480 sec = 8 mins

tendrl/commons/objects/cluster/atoms/setup_cluster_alias/__init__.py
```diff
@@ -35,22 +35,31 @@ def run(self):
...
-                        integration_id
+                        "message": msg
                     },
                     job_id=self.parameters['job_id'],
                     flow_id=self.parameters['flow_id'],
                 )
                 return False
             time.sleep(5)
             finished = True
-            job = Job(job_id=_job_id).load()
```
shtripat: It should be loaded again for job

tendrl/commons/objects/definition/master.yaml
```diff
@@ -938,6 +945,13 @@ namespace.tendrl:
           run: tendrl.objects.Node.atoms.Cmd
           type: Create
           uuid: dc8fff3a-34d9-4786-9282-55eff6abb6c3
+        CheckServiceStatus:
```
shtripat: May be rename as `CheckRequiredServicesRunning`

tendrl/commons/objects/node/atoms/check_service_status/__init__.py
```diff
@@ -0,0 +1,142 @@
...
+                        NS.get("publisher_id", None),
+                        {
+                            "message": msg
+                        },
+                        job_id=self.parameters['job_id'],
+                        flow_id=self.parameters['flow_id']
+                    )
+                    service_status = False
+            node_list = self.parameters['Node[]']
+            if service_status and len(node_list) > 1:
```
shtripat: Add a comment sating that the monitoring integration service check should happen from the bootstrap node only

tendrl/commons/objects/node/atoms/check_service_status/__init__.py
```diff
@@ -0,0 +1,142 @@
...
+                try:
+                    node_arr = etcd_utils.read(
+                        "/indexes/tags/tendrl/integration/monitoring"
+                    ).value
+                    node_id = eval(node_arr)[0]
+                    service = NS.tendrl.objects.Service(
+                        service="tendrl-monitoring-integration",
+                        node_id=node_id
+                    ).load()
+                    service.error = list(service.error)
```
shtripat: Check if this is required

tendrl/commons/objects/node/atoms/check_service_status/__init__.py
```diff
@@ -0,0 +1,142 @@
...
+                            "message": msg
+                        },
+                        job_id=self.parameters['job_id'],
+                        flow_id=self.parameters['flow_id']
+                    )
+                    service_status = False
+        return service_status
+
+
+def get_service_status(service_name):
```
shtripat: This should not be part of Atom file

tendrl/commons/objects/cluster/atoms/import_cluster/__init__.py
GowthamShanmugam: ack

tendrl/commons/jobs/__init__.py
GowthamShanmugam: i will do that

tendrl/commons/objects/node/atoms/check_service_status/__init__.py
GowthamShanmugam: i have checked this, we don't need to add this check, because it is already in list type

GowthamShanmugam:
@shtripat please review

tendrl/commons/objects/node/atoms/check_required_services_running/__init__.py
```diff
@@ -0,0 +1,120 @@
...
+                            NS.get("publisher_id", None),
+                            {
+                                "message": msg
+                            },
+                            job_id=self.parameters['job_id'],
+                            flow_id=self.parameters['flow_id']
+                        )
+                        required_services_running = False
+                except (etcd.EtcdKeyNotFound, IndexError):
+                    msg = ("Service tendrl-monitoring-integration is "
```
shtripat: In case of exception, shouldnt we mention that its unable to check the running status of tendrl-monitoring-integration?

GowthamShanmugam: It make sense, i think i can modify the log message like this

tendrl/commons/jobs/__init__.py
shtripat: So this sleep for avoiding clash of threads. Fine.

GowthamShanmugam: > So this sleep for avoiding clash of threads. Fine.

yes, even all nodes are pick up the job at the same time, this sleep will stop the node execution for 3 sec and tell them who has locked the job recently. So only the node which is locked the job can continue the execution.

tendrl/commons/objects/node/atoms/check_required_services_running/__init__.py
```diff
@@ -0,0 +1,119 @@
...
+            )
+        else:
+            # check neccessary service status in storage nodes
+            required_services_running = self.check_service_status(
+                TENDRL_SERVICES["storage_node"],
+                NS.node_context.fqdn
+            )
+            node_list = self.parameters['Node[]']
+            if required_services_running and len(node_list) > 1:
+                # check monitoring integration is running on server from
```
shtripat: What is this piece of code for? Why `tendrl-monitoring-integration` should be running on a storage node ?

tendrl/commons/objects/cluster/atoms/setup_cluster_alias/__init__.py
GowthamShanmugam: yes, it will create and populate the fields in new object

tendrl/commons/objects/node/atoms/check_required_services_running/__init__.py
GowthamShanmugam: This atom is shared by un-manage and import flow, This is a pre-check to figure out whether all necessary services are running for the flow or not. 
For import flow job is processed by storage nodes, So I am checking node-agent, glusterd by executing ansible module and i am checking monitoring-integration status from ETCD.
Because setup alias is a monitoring integration job.  So before creating a job, the storage node will make sure monitoring-integration is running or not.

shtripat: Ack. Looks fine.

PR #1083, opened 2019-04-01
Timeout only parent job when it is not processed for long time

GowthamShanmugam:
bugzilla: 1602858
tendrl-bug-id: #1080


GowthamShanmugam:
@shtripat please review

Issue #1084, opened 2019-04-24
Import Cluster fails on "node X doesn't have network details populated" with Ansible 2.8


GowthamShanmugam:
With Beta 1 build of Ansible 2.8, it's not possible to import Gluster Trusted
Storage pool into RHGSWA, as Cluster Import task fails with error:

 Node <X> doesn't have network details populated

SalsaBr:
I ran across this same sympton and could not recover from the error downgrading to ansible 2.7 nor trying to unmanage the cluster. The cluster is stuck in an error state.

GowthamShanmugam:
Does /tmp directory have execution permission? if not, please remount /tmp directory with exec permission: https://askubuntu.com/questions/311438/how-to-make-tmp-executable

SalsaBr:
Yes, it does. On all nodes

GowthamShanmugam:
Please unmanage the cluster and wait for all the nodes will be detected by tendrl server. Fire import after all the nodes are listed with fqdn.

SalsaBr:
The unmanage function is not working either. Nothing happens and I can't check on it's progress.

GowthamShanmugam:
Oh ok you already mentioned like un-manage is not working sorry :),  Please check the log file in /var/log/messages. I think in each sync it may populate the error in the log file.

GowthamShanmugam:
If nothing working I will help you in a remote call to solve this problem

SalsaBr:
I tried to clear everything by deleting my etcd partition, then installed tendrl againand this is what I got now:

Child jobs failed are [u'02239281-7b28-4cec-8d11-7c9f6b82fca1']

Failed atom: tendrl.objects.Cluster.atoms.ImportCluster on flow: Import existing Gluster Cluster

[...]

PR #1085, opened 2019-04-24
Import Cluster fails on "node X doesn't have network details populated" with Ansible 2.8

GowthamShanmugam:
bugzilla: 1702412
tendrl-bug-id: Tendrl/commons#1084

### DIFF
--- a/tendrl/commons/utils/ansible_module_runner.py
+++ b/tendrl/commons/utils/ansible_module_runner.py
@@ -1,9 +1,12 @@
 import os
 import subprocess
+import sys
 import tempfile
 
 import ansible.executor.module_common as module_common
 from ansible import modules
+from ansible.parsing.dataloader import DataLoader
+from ansible.template import Templar
 
 from tendrl.commons.utils import log_utils as logger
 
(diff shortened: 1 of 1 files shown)


GowthamShanmugam:
@shtripat @nthomas-redhat This solution will work in current supporting ansible version as well as ansible 2.8.

tendrl/commons/utils/ansible_module_runner.py
```diff
@@ -1,9 +1,12 @@
 import os
 import subprocess
+import sys
 import tempfile
 
 import ansible.executor.module_common as module_common
 from ansible import modules
+from ansible.parsing.dataloader import DataLoader
+from ansible.template import Templar
```
shtripat: Should the import variable be called `Templar` or `Template` ?

GowthamShanmugam: it is templar only @shtripat https://github.com/ansible/ansible/blob/4d3a6123d5b84ed8cdaba4e945e6f1d55c35c4b4/lib/ansible/template/__init__.py#L267

shtripat: Ack.

PR #1091, opened 2020-03-16
Incorrect usage is corrected for global variable

TimothyAsirJeyasing:
Incorrect usage of global variable is corrected
Spec file changes for python3 conversion

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,7 +7,7 @@ from setuptools import setup
 
 try:
     # Python 2 backwards compat
-    from __builtin__ import raw_input as input
+    from builtins import raw_input as input
 except ImportError:
     pass
 
--- a/setup.py
+++ b/setup.py
@@ -7,7 +7,7 @@ from setuptools import setup
 
 try:
     # Python 2 backwards compat
-    from builtins import raw_input as input
+    from builtins import input
 except ImportError:
     pass
 
--- a/tendrl/commons/objects/cluster/atoms/import_cluster/__init__.py
+++ b/tendrl/commons/objects/cluster/atoms/import_cluster/__init__.py
@@ -42,7 +42,7 @@ class ImportCluster(objects.BaseAtom):
                 cmd = cmd_utils.Command('gluster volume list')
                 out, err, rc = cmd.run()
                 if not err:
-                    volumes = filter(None, out.split("\n"))
+                    volumes = [_f for _f in out.split("\n") if _f]
                     ret_val, err = enable_disable_volume_profiling(
                         volumes, self.parameters)
                     if not ret_val:
--- a/tendrl/commons/objects/cluster/atoms/import_cluster/__init__.py
+++ b/tendrl/commons/objects/cluster/atoms/import_cluster/__init__.py
@@ -1,6 +1,7 @@
 import subprocess
 import time
 import uuid
+import os
 
 from tendrl.commons.event import Event
 from tendrl.commons.flows.import_cluster.gluster_help \
@@ -42,7 +43,7 @@ class ImportCluster(objects.BaseAtom):
                 cmd = cmd_utils.Command('gluster volume list')
                 out, err, rc = cmd.run()
                 if not err:
-                    volumes = [_f for _f in out.split("\n") if _f]
+                    volumes = [_f for _f in out.split(os.linesep) if _f]
                     ret_val, err = enable_disable_volume_profiling(
                         volumes, self.parameters)
                     if not ret_val:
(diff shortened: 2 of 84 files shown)


TimothyAsirJeyasing:
@amohan@redhat.com, @nthomas-redhat, @anmolbabu , @cloudbehl  Please review

cloudbehl:
@TimothyAsirJeyasing can you fix the `Codacy/PR Quality Review` and also why CI tests are failing?

tendrl/commons/tests/flows/expand_cluster/test_ec_init.py
```diff
@@ -1,4 +1,4 @@
-import __builtin__
+import builtins
```
cloudbehl, line 1: use builtins instead of __builtin__ in code also. Check line 50

cloudbehl (changes requested):
Please make this change in all the test files.

tendrl/commons/tests/flows/expand_cluster/test_ec_init.py
TimothyAsirJeyasing: Done

setup.py
```diff
@@ -7,7 +7,7 @@
 
 try:
     # Python 2 backwards compat
-    from __builtin__ import raw_input as input
+    from builtins import raw_input as input
```
aruniiird: the above line is giving following error in my python3 setup,
```python
>>> from builtins import raw_input as input
  File "<stdin>", line 1, in <module>
ImportError: cannot import name 'raw_input' from 'builtins' (unknown location)
>>> 
```
you may have to change it to,
```python
from six.moves import input
```
this will work in both python 2.7 and 3.x (tested on python 2.7.16 and python 3.7.5)

TimothyAsirJeyasing:
@GowthamShanmugam , @shtripat Please review

tendrl/commons/utils/central_store/utils.py
```diff
@@ -2,7 +2,7 @@
 import datetime
 import etcd
 import sys
-import thread
+import _thread
```
aruniiird, line 5:
```suggestion
import _thread as thread
```
then you don't have to change anywhere else

aruniiird (changes requested):

tendrl/commons/utils/central_store/fields.py
```diff
@@ -244,7 +244,7 @@ def _set_value(self, value):
 
         # Force casting if we were given a caster.
         if self._caster:
-            for x in value.keys():
+            for x in list(value.keys()):
```
GowthamShanmugam: I am seeing these changes in a few places, it is not necessary to convert this list by default it will give the result in list only. 

But not sure please check :)

GowthamShanmugam: no need list(value.keys()) or list((value.values()) or len(list((value.keys())) no need list conversion if i am right

GowthamShanmugam (commented):
only minor suggestion, not change LGTM

setup.py
GowthamShanmugam: can you try this: 
    from builtins import input

no need alias, raw_input is renamed as input

TimothyAsirJeyasing: Thank you, I will do it!

tendrl/commons/utils/central_store/utils.py
TimothyAsirJeyasing: Sure thank you

tendrl/commons/utils/central_store/fields.py
TimothyAsirJeyasing: Done

TimothyAsirJeyasing:
Have done the review changes. Please review @GowthamShanmugam , @aruniiird

tendrl/commons/objects/cluster/atoms/import_cluster/__init__.py
```diff
@@ -42,7 +42,7 @@ def run(self):
                 cmd = cmd_utils.Command('gluster volume list')
                 out, err, rc = cmd.run()
                 if not err:
-                    volumes = filter(None, out.split("\n"))
+                    volumes = [_f for _f in out.split("\n") if _f]
```
aruniiird: Instead of "\n" , isn't it a good idea to use, `os.linesep` (which will give more interoperability), but leaving this at your discretion.

aruniiird (commented):
A minor comment, nothing much to review here...

tendrl/commons/objects/cluster/atoms/import_cluster/__init__.py
TimothyAsirJeyasing: This patch is mainly for python3 convert. However we can do this kind of enhancement at later point of time. Thank you.

